PR #20, opened 2018-12-11
Fix-reference-box

kpwilson:
removed image link, but not image.


kpwilson:
@sidwray - fixing images. i don't have permission to edit directly in gh-pages. i need an approver.

PR #25, opened 2018-12-11
Full Dev to Master PR

sidwray:
I think this is the full merge of Dev to Master for all the recent changes.  If there are unexpected commits that would happen we can decline it.


sidwray:
Canceling to re-start

PR #49, opened 2018-12-20
Fix missing link in README

lpetschauer:
The `Why Connectors` link in our README's table of contents doesn't have a target and 404s when clicked.


sidwray:
Thanks @lpetschauer! Asking Kathy to review.

PR #57, opened 2019-01-03
Update share.md

kpwilson:
Formatted content as numbered steps. Added "--force-keys" information.

### DIFF
--- a/docs/share.md
+++ b/docs/share.md
@@ -12,19 +12,28 @@ You can share the data your connector accesses by connecting to the data, creati
 
 Or, if you prefer to have your users test the connector, you can place the connector files in the directories and then run the commands as follows:
 
-**Tableau Desktop:** This is the same as developing the connector.
-Place the files into a subdirectory you create, and then use the -DConnectPluginsPath command line argument.  For example, place your plugin files in `C:\tableau_connectors\myconnector` and then start Tableau:
+**Tableau Desktop:** 
+
+1. Place the files into a subdirectory you create.
+1. Use the `-DConnectPluginsPath` command line argument.  
+
+For example, place your connector files in `C:\tableau_connectors\myconnector` and then start Tableau:
 
 ```
 tableau.exe -DConnectPluginsPath=C:\tableau_connectors
 ```
 
-**Tableau Server:** Follow the same instructions as Tableau Desktop for each server node.
-Then set this option: "native_api.connect_plugins_path: C:/tableau_connectors".
-For information about using TSM to do this, see [tsm configuration set Options](https://onlinehelp.tableau.com/current/server-linux/en-us/cli_configuration-set_tsm.htm) in the Tableau Server help.
+**Tableau Server:** 
+
+1. Follow the same instructions as Tableau Desktop for each server node.
+1. Set this option: `native_api.connect_plugins_path: C:/tableau_connectors`
+
+For information about using TSM to set the option, see [tsm configuration set Options](https://onlinehelp.tableau.com/current/server-linux/en-us/cli_configuration-set_tsm.htm) in the Tableau Server help.
 
 For example:
 
 ```
-tsm configuration set -k native_api.connect_plugins_path -v C:/tableau_connectors
+tsm configuration set -k native_api.connect_plugins_path -v C:/tableau_connectors 
 ```
+
+If you get a configuration error, try adding the `--force-keys` parameter to the command.


docs/share.md
```diff
@@ -12,19 +12,28 @@ You can share the data your connector accesses by connecting to the data, creati
 
 Or, if you prefer to have your users test the connector, you can place the connector files in the directories and then run the commands as follows:
 
-**Tableau Desktop:** This is the same as developing the connector.
-Place the files into a subdirectory you create, and then use the -DConnectPluginsPath command line argument.  For example, place your plugin files in `C:\tableau_connectors\myconnector` and then start Tableau:
+**Tableau Desktop:** 
+
+1. Place the files into a subdirectory you create.
```
pvanderknyff: I think we're losing some information here. The line "For example, place your plugin files in `C:\tableau_connectors\myconnector`" is removed, and now it's a bit more vague. The problem here was some confusion over where the command line argument should point to. My suggestion (wording may not be very good):

1. Create a tableau connector directory (For example, C:\tableau_connectors)
2. Place the folder containing your connector's manifest.xml file in this directory. Each connector should have its own folder. (For example, C:\tableau_connectors\my_connector)
3. Run tableau using the -DConnectPluginsPath command line argument, pointing to your connector directory. (For example, -DConnectPluginsPath=C:\tableau_connectors

docs/share.md
```diff
@@ -12,19 +12,28 @@ You can share the data your connector accesses by connecting to the data, creati
...
+1. Use the `-DConnectPluginsPath` command line argument.
```
pvanderknyff: Should be 2, not 1.

docs/share.md
```diff
@@ -12,19 +12,28 @@ You can share the data your connector accesses by connecting to the data, creati
...
+For information about using TSM to set the option, see [tsm configuration set Options](https://onlinehelp.tableau.com/current/server-linux/en-us/cli_configuration-set_tsm.htm) in the Tableau Server help.
 
 For example:
 
 ```
-tsm configuration set -k native_api.connect_plugins_path -v C:/tableau_connectors
+tsm configuration set -k native_api.connect_plugins_path -v C:/tableau_connectors 
 ```
+
+If you get a configuration error, try adding the `--force-keys` option to the end of the command.
```
pvanderknyff: Not sure where, but we may want to mention that for the server to see changes to the connectors, it must be restarted. For example, if you drag and drop a new connector into the server's connector directory, it will not automatically show up until the server is restarted. I can see that maybe confusing some people wondering why their brand new connector isn't showing up.

docs/share.md
sidwray: Great suggestions, thanks!

kpwilson: +1 Incorporated changes.

docs/share.md
kpwilson: That's the markdown numbering scheme. It displays as 2.

docs/share.md
kpwilson: I added Restart server as the last step in the Tableau Server section.

docs/share.md
```diff
@@ -12,19 +12,28 @@ You can share the data your connector accesses by connecting to the data, creati
...
-tsm configuration set -k native_api.connect_plugins_path -v C:/tableau_connectors
-```
+1. For each server node, follow Tableau Desktop's steps 1 and 2 above.
+1. Set the `native_api.connect_plugins_path` option. For example:
+
+    ```
+    tsm configuration set -k native_api.connect_plugins_path -v C:/tableau_connectors 
+    ```
+  
+1. Restart Tableau Server to see the new connector.
```
pvanderknyff: Looks pretty good! Just one last thing.

It's mentioned in the linked tsm help page, but to apply the changes you need to run the proper command, which automatically restarts the server. This isn't a tutorial on tsm, and we don't want it to be, so walking the reader through applying the change isn't necessarily. The upshot is that for step 2, they will have already restarted the server, so another restart is redundant.

My initial comment on restarting the server was imagining it as a comment outside of the steps, since it will come up not when someone is initially trying to get their connectors working on a server but when they update their connectors and want to see their changes reflected on the server. Maybe something like:
"Updating, removing or adding connectors will require a server restart before the changes show up on the server."

kpwilson: Got it. I removed the step and added a note about restarting the server when changes are made.

PR #58, opened 2019-01-07
Add section about web editing

lriggs:


lriggs:
I changed "Editing" to "Authoring" since that seems to be the accepted term.

PR #60, opened 2019-02-06
Added new dialect (TDD) topic. Updated TOC and Design with new topic.

kpwilson:
Added new TDD topic.


kpwilson:
Added TDD topic.

PR #62, opened 2019-02-07
TFS-852745 TDVT: Better error messages when running from unexpected directories

lukewrites:
This PR gives a detailed error message if there are no data sources found in the directory where `TDVT` is run rather than throwing a `TypeError` as it does now.

Additionally, I made some little edits to the file to get rid of PEP8 warnings and to make the formatting of the "available suites" list a little prettier.


lukewrites:
@lriggs The change in wording is at `tdvt/tdvt/config_gen/datasource_list.py` line 16; I pulled it out into a constant to make it easier to reword and to use in the tests.

PR #68, opened 2019-02-28
Dev

lriggs:
Example of adding two arbitrary text boxes to the connection dialog and using the values in the connection builder.


samples/components/dialogs/new text field/connection-dialog.tcd
```diff
@@ -0,0 +1,19 @@
+<connection-dialog class='kyvos'>
```
pvanderknyff: Here, the class is 'kyvos', and below, in the tdr it's 'postgres_odbc2'. Even if this isn't a full plugin (I don't see a manifest), they should probably be aligned. Maybe use 'new_text_field_sample' to be descriptive?

lriggs: Thanks good catch.

lriggs: Interestingly it worked like this.

lriggs:
> 
> 
> Just for clarity's sake on my part, this sample using existing text boxes in new ways, correct?

Yes.

rosswbrown:
Service is a great example of an arbitrary extension point in the dialog.  Warehouse gets a little trickier longer term.  We likely in the future will want to allow HasWarehouses to be true in the tcd as a part of more flexible metadata enumeration, potentially giving that attribute meaning more globally in the platform.  No objections though.

samples/components/dialogs/new text field/connectionRequired.js
```diff
@@ -0,0 +1,4 @@
+(function requiredAttrs(attr)
+{
+    return ["class", "server", "port", "dbname", "username", "password", connectionHelper.attributeWarehouse, "service"];
```
rosswbrown: more consistent usage of connectionHelper in samples when possible

PR #71, opened 2019-03-07
Remove math on None

lukewrites:
At the moment our code will attempt to perform subtraction on `len(None)`, which throws a TypeError. This is a small PR to (a) stop tdvt from trying to find the length of `None` or do math on `None`, while giving good error messages.

Screen Shot 2019-03-08 at 12 05 12 PM


lukewrites:
My editor killed a bunch of whitespace, so the diff is ugly. 😬

PR #79, opened 2019-03-14
update formatKeyValuePair, log, locateDriver function names

jsun28:
update formatKeyValuePair, log, locateDriver function names


pvanderknyff:
We should probably wait until after 2019.2, since the functions won't be updated until then.

lukewrites:
@pvanderknyff in that case it sounds like we should consider making a holding branch for `2019.2` related docs.

pvanderknyff:
LGTM.

PR #82, opened 2019-03-22
Add vendor information info to Examples page

pvanderknyff:
Updated the examples page to document the new company name and support link requirements in the manifest file.

### DIFF
--- a/docs/example.md
+++ b/docs/example.md
@@ -17,7 +17,7 @@ The <span style="color:red">class</span> value is a unique key for your connecto
 Each connector is typically based on a "class" such as ODBC or JDBC, and provides additional customizations beyond the class.
 The <span style="color:blue; font-family: courier new">name</span> value displays the connector name in the Tableau **Connect** pane.
 
-The vendor information must also be specified. The <span style="color:blue; font-family: courier new">company name</span> value displays the name of the creator of the connector in the Tableau **Connect** pane, and the <span style="color:blue; font-family: courier new">support link</span> value is the url of a website that users of the connector can go to to get support.
+You must also specify the vendor information. The <span style="font-family: courier new">company name</span> value displays the name of the creator of the connector in the Tableau **Connect** pane, and the <span style="font-family: courier new">support link</span> value is the URL of a website where users of the connector can get support.
 
 ![]({{ site.baseurl }}/assets/manifest-xml.png)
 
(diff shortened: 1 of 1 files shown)


docs/example.md
```diff
@@ -17,6 +17,8 @@ The <span style="color:red">class</span> value is a unique key for your connecto
 Each connector is typically based on a "class" such as ODBC or JDBC, and provides additional customizations beyond the class.
 The <span style="color:blue; font-family: courier new">name</span> value displays the connector name in the Tableau **Connect** pane.
 
+The vendor information must also be specified. The <span style="color:blue; font-family: courier new">company name</span> value displays the name of the creator of the connector in the Tableau **Connect** pane, and the <span style="color:blue; font-family: courier new">support link</span> value is the url of a website that users of the connector can go to to get support.
```
kpwilson:
```suggestion
You must also specify the vendor information. The <span style="font-family: courier new">company name</span> value displays the name of the creator of the connector in the Tableau **Connect** pane, and the <span style="font-family: courier new">support link</span> value is the URL of a website where users of the connector can get support.
```
I'd suggest putting this paragraph under the little connect pane image.
Questions: 
Is the name of the creator going to display on the Connect pane, along with the name of the connector? It's not clear with the current sentence. 
Where will the support link display?

kpwilson: That second sentence could say:
The company name value displays the name of the connector's creator next to the connector name on the Tableau Connect pane (for example, MyCompany MyConnector <or some other brilliant example>,

pvanderknyff: The name of the creator will be displayed in the next to the connector name like so: Connector Name by Creator.

The support link question is trickier, since it is not displaying anywhere at the moment. We're at the awkward moment between it being required in the manifest but before it displays in the UI.

kpwilson:
```suggestion
You must also specify the vendor information. The <span style="font-family: courier new">company name</span> value displays the name of the connector's creator next to the connector name on the Tableau **Connect** pane (for example, Connector Name by Creator), and the <span style="color:blue; font-family: courier new">support link</span> value is the URL of a website where users of the connector can get support.
```
Is that accurate?

pvanderknyff: I think we call it Company Name in the xml, not Creator, so we should probably keep that terminology going. Other than that, it looks accurate.

pvanderknyff: Nevermind, it makes sense in context.

kpwilson: Sounds good. When we know where the support link will display, it'd be nice to add that info here.

PR #85, opened 2019-03-28
Create stub for XSD Validation

pvanderknyff:
Initial stub. I added the XSD validator as a separate file, and used TDVT as a guide for setting that up to put that in a module we can import from.

### DIFF
--- a/packaging/packaging/xsd_validator.py
+++ b/packaging/packaging/xsd_validator.py
@@ -1,0 +1,20 @@
+
+# files_list: List of files to validate
+# folder_path: path to folder that contains the files.
+# Return value: True if all xml files pass validation, false if they do not or there is an error
+def validate_xsd(files_list, folder_path = ""):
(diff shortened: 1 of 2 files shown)


packaging/packaging/xsd_validator.py
```diff
@@ -0,0 +1,20 @@
+
+# files_list: List of files to validate
+# folder_path: path to folder that contains the files.
+# Return value: True if all xml files pass validation, false if they do not or there is an error
+def validate_xsd(files_list, folder_path = ""):
+    print("===Start XSD Validation===")
```
lriggs: I'd suggest we use the same logging mechanism as TDVT, This is ok for a stub, but for future reference:
https://github.com/tableau/connector-plugin-sdk/blob/master/tdvt/tdvt/tdvt.py#L409

You can give it different levels and print to the console and/or a log file.

PR #90, opened 2019-04-01
flake8 nits plus a docstring

lukewrites:
* Fixed some flake8 nits (you can [have VS Code do it for you](https://code.visualstudio.com/docs/python/linting)).
* Added a [docstring](
https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) that may not be 100% PEP-257 friendly because it used [this VS Code plugin](https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring)

### DIFF
--- a/packaging/package/xsd_validator.py
+++ b/packaging/package/xsd_validator.py
@@ -1,12 +1,14 @@
 import sys
-import xmlschema
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


lukewrites:
@pvanderknyff you can add the `summary` to the function if you want 😃

PR #92, opened 2019-04-02
 Dev workflow, venv, build, test, instructions

rosswbrown:
See README for supported scenarios.  Let me know if anything isn't best practice.  Other option is check-in and iterate some more.  Will address renames in a future change.


rosswbrown:
> What are your thoughts behind recommending that users install in a virtual environment?

I was trying to consider this in part if we continued to use fairly generic names.  If we go to something more specific recommending for users to install global package might be fine.

Also not sure if there's a common bootstrap technique that does venv more automatically for users.

lukewrites:
Missed getting to make this comment on the PR, but do we need a `version.py` file to go with `__version__` in `setup.py`?

PR #94, opened 2019-04-03
Package jsun

jsun28:
Add jar package functionality, last PR was discarded.
manifest package refactored, add flexibility to define jar file destination dir.

### DIFF
--- a/packaging/package/jar_packager.py
+++ b/packaging/package/jar_packager.py
@@ -28,6 +28,10 @@ def create_jar(source_dir, files, jar_filename, dest_dir="..\jar"):
     abs_source_path = os.path.abspath(source_dir)
     print("Start packaging", jar_filename, "from", abs_source_path)
 
+    # if dest dir doesn't exist, then create it
+    if not os.path.exists(dest_dir):
+        os.makedirs(dest_dir)
+
     with ZipFile(os.path.join(dest_dir, Path(jar_filename)), "w") as jar:
         jar.writestr("META-INF/", "")
         jar.writestr("META-INF/MANIFEST.MF", Manifest().get_data())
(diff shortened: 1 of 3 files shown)


packaging/package/package.py
```diff
@@ -20,7 +21,10 @@ def main():
         ConnectorFile("connectionResolver.tdr", "connection-resolver")]
 
     validate_xsd(files_to_package, path_from_args)
```
pvanderknyff: I think we should make the packaging only happen if validate_xsd returns true. Otherwise, we may try to package a plugin we know is broken.

packaging/package/jar_packager.py
```diff
@@ -0,0 +1,57 @@
+import os
+
+from connector_file import ConnectorFile
+from zipfile import ZipFile
```
pvanderknyff: Did you have to pip install any of these modules? If so, it needs to be added to setup.y's install field.

lukewrites: `os`, `zipfile`, and `pathlib` are part of the standard library, so they're gtg 👍 

I would suggest reordering the imports [per PEP-8](https://www.python.org/dev/peps/pep-0008/#imports): standard library, 3rd party, local applications.

```python
import os
from pathlib import PathLib
from zipfile import ZipFile

from connector_file import ConnectorFile
from manifest import Manifest
```

packaging/package/package.py
jsun28: ok sounds good.

packaging/package/jar_packager.py
jsun28: ok I will merge this PR and send another PR about those two issues, since it takes a while to revert remote branch and recommit the whole change.

PR #96, opened 2019-04-03
Fix so that all modules run from same location

pvanderknyff:
According to the readme, we should be running this from connector-plugin-sdk/packaging, but package.py and jar_packager.py would not find the imports if run from there.

### DIFF
--- a/packaging/package/jar_packager.py
+++ b/packaging/package/jar_packager.py
@@ -1,8 +1,8 @@
 import os
 
-from connector_file import ConnectorFile
+from .connector_file import ConnectorFile
 from zipfile import ZipFile
-from manifest import Manifest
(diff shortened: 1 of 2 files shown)


jsun28:
looks good, thanks for also take relative path fixing.

lukewrites:
I wonder if it would be a little more readable to use absolute imports? (`from package.manifest import Manifest` etc) Works either way, just whatever standard we want to use 😄

pvanderknyff:
> I wonder if it would be a little more readable to use absolute imports? (`from package.manifest import Manifest` etc) Works either way, just whatever standard we want to use 😄

I wasn't sure if Jianan was working on these files, and wanted it to be as simple a merge as possible.

PR #97, opened 2019-04-03
add six to install_requires

lukewrites:
`six` is imported in `manifest.py`


lukewrites:
When I did an install in a virtualenv that was the only thing that threw an error when I tried to do `python -m package.package`.

PR #98, opened 2019-04-03
Add logging instead of print statements

pvanderknyff:
Changes most print statements to logging.debug. There are a few I made warnings. This means that, unless verbose is set to true via a command line argument, we do not print it to the console. (It's hardcoded right now since we haven't set that up yet.) We will, however, always save it all log messages to packaging_log.txt. 

Let me know if there are some messages that should be shown to the console that I have labeled as debug.

### DIFF
--- a/packaging/package/jar_packager.py
+++ b/packaging/package/jar_packager.py
@@ -1,10 +1,13 @@
 import os
+import logging
 
 from pathlib import Path
 from zipfile import ZipFile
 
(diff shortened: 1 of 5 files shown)


pvanderknyff:
Added logic to the __init__.py file for the tests folder to deal with logging. Made it so it will not print anything during the unit tests unless it's at log level error or higher, but it will save the logs in a text file.

PR #99, opened 2019-04-03
change validate xml then create jar logic, and change default jar file destination

jsun28:
1. update validate XML then create jar logic in package.py
2. change relative path, default jar file destination dir will be connector-plugin-sdk\packaging\jar


jsun28:
OK I'll merge it now

PR #104, opened 2019-04-04
delete default dest dir

jsun28:


pvanderknyff:
I think we should keep this in for now. Maybe change the name, but specifying the default directory each time is a bit unwieldy.

jsun28:
yeah agree, specifying the default directory each time might be a bit annoying.

PR #107, opened 2019-04-05
Plugin File List Generator + Fix Validation Bug

pvanderknyff:
First version is working of generating list of files to package. Things left to do:
- Make sure that we can't get stuck in an infinite loop if two files link to each other
- Also pick up and add resource files.

Also fixed Defect 894385, since it was easy and making it hard to test this part.


pvanderknyff:
Having some trouble with an edge case, will close this and re-open a new one since I'm going to be changing a bit.

pvanderknyff:
Refactored the file generated logic as part of an XMLParser class. By making it a class I made the method argument lists much cleaner.

packaging/package/package.py
```diff
@@ -45,46 +46,39 @@ def init():
...
-        ConnectorFile("connectionResolver.tdr", "connection-resolver")]
     
     if args.package:
+               
         path_from_args = Path(args.package)
 
-        if not path_from_args.is_dir():
-            logger.warning("Error: " + str(path_from_args) + " does not exist or is not a directory.")   
-            return  
+        parser = XMLParser(path_from_args)
```
rosswbrown: reusing variable name parser, argparser/xmlparser?

pvanderknyff: Fixed, thanks!

PR #108, opened 2019-04-05
refactor manifest and add vendor section

jsun28:
1. refactor manifest package about sections
2. add vendor section in generated manifest file, new section looks like:
Created-By: Tableau Connector SDK 0.0.1

### DIFF
--- a/packaging/package/manifest.py
+++ b/packaging/package/manifest.py
@@ -6,8 +6,12 @@ import os
 
 from collections import OrderedDict
 from six import BytesIO
+from .version import __version__
 
 manifest_line_length = 70
+MANIFEST_VERSION = "Manifest-Version"
+VENDOR_KEY = "Created-By"
+SDK_NAME = "Tableau Connector SDK "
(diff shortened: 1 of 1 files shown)


packaging/package/manifest.py
```diff
@@ -6,8 +6,12 @@
 
 from collections import OrderedDict
 from six import BytesIO
+from .version import __version__
 
 manifest_line_length = 70
+MANIFEST_VERSION = "Manifest-Version"
+VENDOR_KEY = "Created-By"
+SDK_NAME = "Tableau Connector SDK "
```
lukewrites: Should this be the SDK or something like `Tableau Packaging Tool`? I'm asking because your example shows the version number of `packaging`, not the SDK (which uses dates in its changelog).

jsun28: yeah, you are right.

jsun28: added a new commit to this PR.

PR #109, opened 2019-04-05
add jar_packager unit test

jsun28:
add an initial version of jar_packager unit test


packaging/tests/test_jar_packager.py
```diff
@@ -0,0 +1,30 @@
+import unittest
+import os.path
+from pathlib import Path
+
+from package.jar_packager import create_jar
+from package.connector_file import ConnectorFile
+
+XSD_TEST_FOLDER = Path("tests/test_resources")
```
pvanderknyff: I think it should be just TEST_FOLDER.

pvanderknyff:
Let's use the same valid connector for both cases. We don't need to have two versions of the same files.

jsun28:
ok cool, will change it

jsun28:
commits are a bit confusing, but file changed should be clear to look at

PR #113, opened 2019-04-08
Package jsun

jsun28:
Add jar_jdk_packager package and call it in package.py

### DIFF
--- a/packaging/package/jar_jdk_packager.py
+++ b/packaging/package/jar_jdk_packager.py
@@ -1,0 +1,53 @@
+import os
+import logging
+import subprocess
+import shutil
+
+from pathlib import Path
+from .connector_file import ConnectorFile
+
+logger = logging.getLogger(__name__)
+
+
+def jdk_create_jar(source_dir, files, jar_filename, dest_dir):
+    """
+    Package JAR file from given files using JAVA JDK
+
+    :param source_dir: source dir of files to be packaged
+    :type source_dir: str
+
+    :param files: files need to be packaged
+    :type files: list of ConnectorFile
+
+    :param jar_filename: filename of the created JAR
+    :type jar_filename: str
+
+    :param dest_dir: destination dir to create jar file
+    :type dest_dir: str
+
+    :return: None
+    """
+    abs_source_path = os.path.abspath(source_dir)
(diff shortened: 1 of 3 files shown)


pvanderknyff:
Is this replacing the current jar_packager.py? If so, we should probably delete the old one. Also make sure to update the unit tests.

For the rest, it looks good to me, but I'm not very knowledgeable on jar packaging.

jsun28:
well, my opinion is to keep the old one in case of we want to use it in the future, especially when we get into the signing part.

rosswbrown:
Is the assumption 'jar' command is on the PATH?  Any assertion of this we can add temporarily?

packaging/package/package.py
```diff
@@ -13,6 +11,7 @@
 LOG_FILE = 'packaging_log.txt'
 PACKAGED_EXTENSION = ".taco"
 
+
```
pvanderknyff: nit: extra blank line

packaging/package/package.py
```diff
@@ -23,6 +22,7 @@ def create_parser():
 
     return parser
 
+
```
pvanderknyff: nit: extra blank line

packaging/package/package.py
```diff
@@ -87,5 +86,6 @@ def main():
     else:
         argparser.print_help()
 
+
```
pvanderknyff: nit: extra blank line

lukewrites:
> Is the assumption 'jar' command is on the PATH? Any assertion of this we can add temporarily?

@rosswbrown it might be found in/with `os.environ['PATH']`?

jsun28:
add a check_jdk_environ_variable function to check PATH has jdk or not. is that something you want here? @rosswbrown

packaging/package/jar_jdk_packager.py
```diff
@@ -0,0 +1,71 @@
...
+    :type jar_filename: str
+
+    :param dest_dir: destination dir to create jar file
+    :type dest_dir: str
+
+    :return: Boolean
+    """
+
+    if not check_jdk_environ_variable():
+        logger.debug("Error: jdk_create_jar: no jdk set up in PATH environment variable, "
```
lriggs: logger.error()

jsun28: sent a new PR out for this

PR #119, opened 2019-04-12
add .pep8speaks.yml

lukewrites:
Starting [pep8speaks](https://pep8speaks.org) with a pretty light touch. The code inspector will only look at _new_ code - won't go through the whole project and list violations.

Over time we can add error codes to be ignored if it gets annoying.

### DIFF
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -1,0 +1,19 @@
+scanner:
+    linter: flake8
+
+flake8:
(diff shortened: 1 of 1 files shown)


pvanderknyff:
Once the line length is made higher, LGTM.

lukewrites:
I'm going to merge this so we can see how it looks…will revert if it's ugly.

PR #125, opened 2019-04-23
Remove -n name argument

rosswbrown:
Allowing tool to rename at packaging time implies functionality we do not intend to support.  1:1 classAttr:packageFile for now.


pvanderknyff:
I still contend that it since the help message says "name of the packaged connector", it does not imply we are allowing them to change the class name, and we can make that more explicit by saying "name of the packaged connector file". Since we can just rename the file after I don't really mind removing it though.

(Also, hopefully Luke can get the line length fixed on the Pep8 bot, it should be 125 characters no 79.)

PR #153, opened 2019-05-14
Dev

lriggs:
Use expected error message from test_set and not TdvtConfig.


tdvt/tdvt/tdvt.py
```diff
@@ -313,8 +313,8 @@ def enqueue_tests(ds_info, args, suite):
             test_config.run_as_perf = ds_info.run_as_perf
             test_config.tds = test_set.tds_name
             test_config.config_file = test_set.config_name
+
```
pvanderknyff: Did you mean to add a space here?

lriggs: I did just to make it more clear. the first block is setting stuff on the conifg and the last line was adding things to the set.

PR #156, opened 2019-05-17
Gh pages dev

lriggs:


docs/capabilities.md
```diff
@@ -63,7 +63,7 @@ CAP_QUERY_CASE_PROMOTES_CHAR | Set to 'no' if CASE cannot promote character type
 CAP_QUERY_FROM_REQUIRES_ALIAS | Set to 'yes' if the FROM clause must provide an alias for the given table. | &ndash; | &ndash;
 CAP_QUERY_GROUP_ALLOW_DUPLICATES | Set to 'no' if SQL queries cannot contain duplicate expressions in the GROUP BY clause (this is uncommon). | yes | yes
 CAP_QUERY_GROUP_BY_ALIAS | Set to 'yes' if SQL queries with aggregations can reference the grouping columns by their corresponding alias in the SELECT list, e.g. GROUP BY "none_ShipCountry_nk". | no | no 
-CAP_QUERY_GROUP_BY_BOOL | Set to 'yes' if the database can group by a raw boolean. Set to 'no' if boolean should be casted to an INT. | &ndash; | **<span style="color:red">yes</span>**
```
rosswbrown: I think the dash is still more consistent with our other documentation, but in practice you're right, unset means 'no' in this case.

lriggs: Now that you mention it, I'm not sure what dash in the default column means exactly. If a capability isn't set it would be 'no' I think by default.
For recommended the dash makes sense.

PR #168, opened 2019-06-26
TFS 928485 Update Tibco tests in TDVT to target version neutral host (Part 1)

chrisjunlee:
These are the new expected values once the changes to the version neutral host applies.

As for immediate impact: composite cases are disabled at the moment, so these changes would not affect the test runs.


.travis.yml
```diff
@@ -24,24 +24,24 @@ jobs:
      python: 3.4
      script: cd $(bash before_deploy.sh)
      deploy:
-     - provider: pypi
+     - provider: pypi  # dev
```
lukewrites: I'm not sure why changes are showing up on this file...what does your git blame show?

tdvt/tdvt/exprtests/standard/expected.setup.operator.date.10.txt
```diff
@@ -94,13 +94,13 @@
         <value>-10491.333</value>
       </tuple>
       <tuple>
-        <value>-85.428888</value>
+        <value>-85.428889</value>
```
lukewrites: These values have changed because of that rounding issue you talked about in standup, right?

lriggs: Is this file only used by Tibco? It might be safer to add another expected if you aren't sure.

.travis.yml
chrisjunlee: Those are Logan's changes. Not sure why git is showing these as diffs when master should already have them.

chrisjunlee: rebased/merged

PR #174, opened 2019-07-02
Add --verify flag for smoke testing

lukewrites:
Description

1. **Adds a `--verify` flag ([`tdvt/tdvt/tdvt.py:431`](https://github.com/tableau/connector-plugin-sdk/pull/174/files#diff-91b3d48efe019954cbc10082ff789f89R431))** that can be used to run smoke tests for the connection to a data source.
    * _Smoke tests_ are defined as either those tests with `SmokeTest = True` in their test definition or the two default `ConnectionTests` decribed below in 2.
3. **Adds two default smoke tests to generated `.ini` files ([` tdvt/tdvt/setup_env.py:84`](https://github.com/tableau/connector-plugin-sdk/pull/174/files#diff-b9e90154014a3dcc5a52b81a2fd89702R84))** that test connections to `cast_calcs` and `Staples`.
    * The two default smoke tests can be disabled individually by setting `CastCalcsTestEnabled = False` and `StaplesTestEnabled = False` in the `.ini` file. They can both be disabled by removing the section.
    * A new test has been added ([`tdvt/tdvt/exprtests/pretest/staples/setup.staples_connection.txt`](https://github.com/tableau/connector-plugin-sdk/pull/174/files#diff-d3a16a215ed863352a8c32b50e665101R1)) to test connecting to `Staples.*.tds`
2. **Changes the normal test run ([`tdvt/tdvt/tdvt.py:445`](https://github.com/tableau/connector-plugin-sdk/pull/174/files#diff-91b3d48efe019954cbc10082ff789f89R482))** to run each data source's smoke tests before running any other tests.
    * Data sources with one or more failing smoke tests will not have any other tests run.
    * Skipped tests count as failing; logs reflect that they were not run because of failing smoke test(s).
    * A warning is shown if there are no smoke tests for a data source.
4. **Allows for other tests to be enabled/disabled** using the `Enabled = False` flag (e.g. [`tdvt/tdvt/config_gen/datasource_list.py:136`](https://github.com/tableau/connector-plugin-sdk/pull/174/files#diff-eee1ebbaefc4bbff6361ca4e4f5f975aR128)).
    * Disabled tests count as failing; logs reflect that they were disabled.
5. **Update docs**, see #182 

### Run a single data source's smoke tests:
Snag_19817fe1

### Run multiple data sources' smoke tests:
Snag_19825cc4

### Run test suite on multiple data sources, one with failing smoke tests, one disabled (`D`) test, and multiple tests skipped (`S`) because of the failing smoke test:
Snag_19851329

In the above and below, `postgres` has the following in its `.ini`:
```
[ConnectionTest]
StaplesTestEnabled = False
CastCalcsTestEnabled = True
```

### Logs reflecting skipped & disabled tests:
Snag_1986eb54
Snag_19870555


tdvt/tdvt/tdvt.py
```diff
@@ -381,38 +393,67 @@ def usage_text():
...
     return parser
 
+
 def init():
     parser = create_parser()
     args = parser.parse_args()
-    #Create logger.
-    logging.basicConfig(filename='tdvt_log_combined.txt',level=logging.DEBUG, filemode='w', format='%(asctime)s %(message)s')
+    if (args.smoke_test is not None) and (args.ds is not None):
+        args.expression_only = 'ConnectionTest'
```
lriggs: I was hoping we could use the smoke test to verify the connection before running the rest of the tests. One way could be to run the pretest against the calcs and stapes tds files then abort the rest of the tests if either of those fail. It would be good to make this flexible enough that it could support a smoke test against any arbitrary tds file. So maybe it could default to calcs and staples and then let users add more in the ini. 

I think this is still a useful first step so that users can easily verify their connections.

lukewrites: Sure, sounds good! I'll keep this in place and add that functionality as well.

tdvt/tdvt/tdvt.py
```diff
@@ -425,51 +467,84 @@ def init():
...
+
 def active_thread_count(threads):
     active = 0
     for t in threads:
         if t.is_alive():
             active += 1
     return active
 
+
+def test_runner(all_tests, test_queue, max_threads):
```
lukewrites: extracted from `run_tests_impl()` below

tdvt/tdvt/setup_env.py
```diff
@@ -81,14 +87,21 @@ def create_ds_ini_file(name, logical_config):
         ini.write('\n')
         ini.write('[UnionTest]\n')
         ini.write('\n')
+        ini.write('[StaplesDataTest]\n')
```
lriggs: I don't know if we want to include this by default. It's a bit problematic since it doesn't support column renaming and the staples table has spaces and capitalization which some databases don't support.

tdvt/tdvt/tdvt.py
```diff
@@ -346,14 +355,18 @@ def usage_text():
         tdvt_runner --run sqlserver,vertica
         tdvt_runner --run standard
 
+    The 'run' argument can also take the --verify flag to run a connection test against const_calcs.
```
lriggs: 'cast_calcs'

tdvt/tdvt/tdvt.py
```diff
@@ -425,51 +467,84 @@ def init():
...
 
-    failed_tests = 0
-    total_tests = 0
-    for work in all_work:
-        if work.copy_files_and_cleanup():
-            print ("Left temp dir: " + work.temp_dir)
-        failed_tests += work.failed_tests if work.failed_tests else 0
-        total_tests += work.total_tests if work.total_tests else 0
+        if failed_smoke_tests > 0:
+            print("Test run aborted due to smoke test errors. {} smoke test(s) failed. Please check logs for information.".format(failed_smoke_tests))
```
lriggs: This is good for a single test suite like a plugin developer would use, but I think it might cause problems with our internal use case of testing many datasources at once. Can the smoke test be tied to the other test sets for the same suite? that way if postgres fails the smoke test then we only skip the postgres tests and not the mysql ones.

lukewrites: I think this should be addressed in fce7f9c.

Doing a `tdvt --run postgres,athena` with athena broken, the smoke tests failed and the regular tests went on to run.

lukewrites: Snag_45e2bf

tdvt/tdvt/setup_env.py
lukewrites: Good point, I've removed it.

tdvt/tdvt/tdvt_core.py
```diff
@@ -389,7 +416,7 @@ def compare_results(test_name, test_file, full_test_file, work):
     try:
         actual_xml = parse(actual_file).getroot()
         result.add_test_results(actual_xml, actual_file)
-    except ElementTree.ParseError as e:
+    except ParseError as e:
```
lukewrites: This was causing an import error; at the top of the file we have `from defusedxml.ElementTree import parse, ParseError`

tdvt/tdvt/config_gen/datasource_list.py
```diff
@@ -74,6 +74,17 @@ def get_expected_message(config):
     return config.get('ExpectedMessage', '')
 
 
+def get_is_smoke_test(config):
+    return config.get('SmokeTest', 'False') == 'True'
```
chrisjunlee: I'm kind of worried about using string compare vs bools. But I see why it's done.

lukewrites: I agree it's ugly. I could change the setting to look for `yes` and `no`, which at least would avoid the appearance of stringy bools.

sonia9703: I feel that it might be better to change the def name to def is_smoke_test(config).

lukewrites: That's a good point, @sonia0504. I named it that way to keep the names similar to `get_password_file` and `get_expected_msg`. But I'm happy to change it.

tdvt/tdvt/config_gen/datasource_list.py
```diff
@@ -273,7 +312,28 @@ def LoadTest(config, test_dir=get_root_dir()):
...
-                                             sect.get('TestPath', ''), test_dir, get_password_file(sect), get_expected_message(sect))
+                                             sect.get('TestPath', ''), test_dir, get_password_file(sect),
+                                             get_expected_message(sect), get_is_smoke_test(sect),
+                                             get_is_test_enabled(sect), False)
+            except KeyError as e:
+                logging.debug(e)
+                pass
+        # Add smoke tests
+        elif connection_test in section:
+            staples_tds_name = 'Staples.*.tds'.replace('*', config_name)
```
lriggs: Since these use the standard tds files you can just pass in CALCS_TDS or STAPLES_TDS and it does the substitution later.

tdvt/tdvt/test_results.py
```diff
@@ -159,7 +169,16 @@ def parse_default_test_cases(self):
         #Parse the setup file to get the count.
         if not self.test_case_map and self.test_set:
             if self.test_set.is_logical:
-                test_result = TestCaseResult("Not run", 0, "", 0, "Not run", self.error_status, None, self.test_config)
+                if self.test_set.test_is_enabled is False:
```
lriggs: It would be nice if you could combine this if/else section with the one below (191 to 204) into a new function. Not required thought if it's difficult.

tdvt/tdvt/config_gen/datasource_list.py
lukewrites: Whoops! Refactored in 6a7ed14 below.

PR #194, opened 2019-07-11
Update Dialect XSD

lriggs:


validation/tdd_latest.xsd
```diff
@@ -111,7 +143,7 @@
       <xs:element maxOccurs="unbounded" name="formula" type="Formula-CT"/>
       <xs:group ref="Function-G"/>
     </xs:sequence>
-    <xs:attribute name="name" type="xs:string" use="required"/>
+    <xs:attribute name="name" type="DateFunction-ST" use="required"/>
```
pvanderknyff: This won't break existing dialects, correct? Otherwise I think this looks good.

lriggs: It shouldn't since those dialects would be causing a logic assert anyways.

PR #199, opened 2019-07-16
Added notes about temp tables and subqueries

lriggs:


docs/design.md
```diff
@@ -79,3 +77,11 @@ You don’t need a dialect definition file if your connector uses the same SQL d
...
 Capabilities can influence the types of queries that Tableau generates, how metadata is read, how Tableau binds to the drivers result set, and many other aspects.
 Common capabilities and how they are used are described in [Capabilities]({{ site.baseurl }}/docs/capabilities).
+
+## Database Capability Considerations
+### Subqueries and Temp Tables
+
+If your database supports temp tables it is recommended that you enable them through the appropriate [Capabilities]({{ site.baseurl }}/docs/capabilities). The connector will perform a simple check at connection time if the temp table capabilities are set in order to confirm that the user has the ability to create a temp table in the current database environment. If the user does not have permission or the capabilities are disabled then Tableau will attempt to generate an alternative query to retrieve the necessary results. Often these queries will need subqueries and the performance can be poor, particularly with large data sets. If the connector does not support temporary tables or subqueries then Tableau will throw an error and will be unable to proceed.
```
pvanderknyff: Nit: Isn't it the database and not the connector that determines support for temp tables and subqueries? 

I would also add the following sentence  at the end of this paragraph and bold it: 
"Thus, if your database does not support either subquries or temp tables, you will not be able to make a connector to that database for Tableau."
This will make absolute sure people understand the implications of what this paragraph says, and we want them to know as soon as possible.

lriggs: I'll make that more clear. I'm not sure what you mean about database vs connector.
A workaround is an extract only connection.

pvanderknyff: I think you're new change should suffice, especially since there is a workaround, but for posterity let me try and clarify what I was saying:
A connector is the code that Tableau uses to connect to the database. In this context, we can't make any changes to the connector code that will allow temp tables to exist if the database does not support temp tables. (Does that make more sense?)

PR #203, opened 2019-07-18
Add signing functionality

jsun28:
Changes:
1. Add jar_jdk_signer package, and call it from package.py
2. Add a helper package, move package.init_logging function and jar_jdk_packager.check_jdk_environ_variable function to helper.
3. Add a few command arguments for signing use 

Command to do packaging and signing:
py -3 -m package.package E:\Github\testdir\postgres_odbc -a jsun -ks "C:\Program Files\Java\jdk1.8.0_201\jre\lib\security\testdir\jsun.jks"
Command to do packaging only:
py -3 -m package.package E:\Github\testdir\postgres_odbc --package-only

### DIFF
--- a/packaging/package/jar_jdk_packager.py
+++ b/packaging/package/jar_jdk_packager.py
@@ -8,7 +8,6 @@ from .connector_file import ConnectorFile
 from .helper import check_jdk_environ_variable
 
 JAR_EXECUTABLE_NAME = "jar.exe"
-PATH_ENVIRON = "PATH"
 logger = logging.getLogger(__name__)
 
 
@@ -32,8 +31,6 @@ def jdk_create_jar(source_dir, files, jar_filename, dest_dir):
     """
 
     if not check_jdk_environ_variable(JAR_EXECUTABLE_NAME):
-        logger.error("Error: jdk_create_jar: no jdk set up in PATH environment variable, "
-                     "please download JAVA JDK and add it to PATH")
         return False
 
     abs_source_path = os.path.abspath(source_dir)
--- a/packaging/package/jar_jdk_packager.py
+++ b/packaging/package/jar_jdk_packager.py
@@ -5,22 +5,11 @@ import shutil
 
 from pathlib import Path
 from .connector_file import ConnectorFile
+from .helper import check_jdk_environ_variable
 
 JAR_EXECUTABLE_NAME = "jar.exe"
 PATH_ENVIRON = "PATH"
 logger = logging.getLogger(__name__)
-
-
-def check_jdk_environ_variable():
-    """
-    Check if jdk is set up in PATH
-    """
-    path_list = os.environ[PATH_ENVIRON].split(';')
-    for path in path_list:
-        if os.path.isfile(Path(path)/JAR_EXECUTABLE_NAME):
-            return True
-
-    return False
 
 
 def jdk_create_jar(source_dir, files, jar_filename, dest_dir):
@@ -42,7 +31,7 @@ def jdk_create_jar(source_dir, files, jar_filename, dest_dir):
     :return: Boolean
     """
 
-    if not check_jdk_environ_variable():
+    if not check_jdk_environ_variable(JAR_EXECUTABLE_NAME):
         logger.error("Error: jdk_create_jar: no jdk set up in PATH environment variable, "
                      "please download JAVA JDK and add it to PATH")
         return False
--- a/packaging/package/jar_jdk_signer.py
+++ b/packaging/package/jar_jdk_signer.py
@@ -1,0 +1,64 @@
+import os
+import logging
+import subprocess
+from pathlib import Path
+from .helper import check_jdk_environ_variable
+
+JARSIGNER_EXECUTABLE_NAME = "jarsigner.exe"
+logger = logging.getLogger(__name__)
+
+
+def validate_signing_input(input_dir, taco_name, alias, keystore):
+    """
+    Validate signing input
+    """
+
+    if not alias:
+        logger.error("Signing Error: Private key's alias is missing or empty")
+        return False
+
+    if not keystore or not os.path.isfile(Path(keystore)):
+        logger.error("Signing Error: Keystore path is missing or invalid")
+        return False
+
+    if not os.path.isfile(input_dir/taco_name):
+        logger.error("Signing Error: Taco file to be signed has been deleted or doesn't exist")
+        return False
+
+    return True
+
+
+def jdk_sign_jar(input_dir, taco_name, alias, keystore):
+    """
+    Sign a taco using JAVA JDK
+
+    :param input_dir: source dir of taco file to be signed
+    :type input_dir: str
+
+    :param taco_name: taco file name
+    :type taco_name: str
+
+    :param alias: Private key's alias in keystore
+    :type alias: str
+
+    :param keystore: keystore path
+    :type keystore: str
+
+    :return: Boolean
+    """
+
+    if not check_jdk_environ_variable(JARSIGNER_EXECUTABLE_NAME):
+        logger.error("Error: jdk_create_jar: no jdk set up in PATH environment variable, "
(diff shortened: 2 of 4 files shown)


packaging/package/package.py
```diff
@@ -57,13 +42,30 @@ def main():
         return
 
     if not files_to_package:
-        logger.info("Packaging failed. Check " + args.log_path + " for more information.")
+        logger.info("Packaging failed due to Connector Files not valid. Check " + args.log_path + " for more information.")
```
pvanderknyff: Wording is a bit awkward, I think it is fine the way it was since packaging and signing are two separate steps.

packaging/package/jar_jdk_signer.py
```diff
@@ -0,0 +1,64 @@
...
+
+    logging.debug("Start signing " + taco_name + " from " + str(os.path.abspath(input_dir)) + " using JDK jarsigner")
+
+    args = ["jarsigner", "-keystore", keystore, "-signedjar", str(input_dir/("signed_" + taco_name)), str(input_dir/taco_name), alias]
+    p = subprocess.Popen(args)
+    p.wait()
+
+    logging.info(taco_name + " was signed as " + "signed_" + taco_name + " at " + str(os.path.abspath(input_dir)))
+
+    return True
```
pvanderknyff: This seems to always return true, even if p failed. What happens if there is an error during the signing subprocess, does it crash the rest of the program? Can we capture and handle the error somehow?

jsun28: good point, I'll try.

packaging/package/package.py
jsun28: done

packaging/package/jar_jdk_signer.py
jsun28: Done, using the exit code to verify the subprocess call result.
0 means succeed, 1 and -1 means failed and unexpected.

packaging/package/jar_jdk_signer.py
```diff
@@ -0,0 +1,66 @@
...
+    if not alias:
+        logger.error("Signing Error: Private key's alias is missing or empty")
+        return False
+
+    if not keystore or not os.path.isfile(Path(keystore)):
+        logger.error("Signing Error: Keystore path is missing or invalid")
+        return False
+
+    if not os.path.isfile(input_dir/taco_name):
+        logger.error("Signing Error: Taco file to be signed has been deleted or doesn't exist")
```
jainam1995: do you also want to check that the extension is .taco ?

packaging/package/jar_jdk_signer.py
```diff
@@ -0,0 +1,66 @@
...
+
+    args = ["jarsigner", "-keystore", keystore, "-signedjar", str(input_dir/("signed_" + taco_name)), str(input_dir/taco_name), alias]
+    p = subprocess.Popen(args)
+    p.wait()
+
+    if p.returncode == 0:
+        logger.info(taco_name + " was signed as " + "signed_" + taco_name + " at " + str(os.path.abspath(input_dir)))
+        return True
+    else:
+        return False
```
jainam1995: For the debug mode, do you also want to output the error message from the jarsigner api to the plugin developer for more information ?

packaging/package/jar_jdk_signer.py
jsun28: Since we can only do package and then sign, the input for signing is the output of 
packaging. see from package_name variable line 49 to line 62 in package.py.

packaging/package/jar_jdk_signer.py
jsun28: I think that's likely fine since the error message from the jarsigner api is already printed out on console.
I tried to use subclass.communicate to get the console out like 
process = subprocess.Popen(args, stdout=subprocess.PIPE)
but one awkward issue is that the when jarsigner pop up the Keystore password, the password user typed in is not hidden.

rosswbrown: at minimum make sure there is a TODO so we resolve that password issue before release

packaging/package/jar_jdk_signer.py
```diff
@@ -0,0 +1,66 @@
...
+
+    logger.debug("Start signing " + taco_name + " from " + str(os.path.abspath(input_dir)) + " using JDK jarsigner")
+
+    args = ["jarsigner", "-keystore", keystore, "-signedjar", str(input_dir/("signed_" + taco_name)), str(input_dir/taco_name), alias]
+    p = subprocess.Popen(args)
+    p.wait()
+
+    if p.returncode == 0:
+        logger.info(taco_name + " was signed as " + "signed_" + taco_name + " at " + str(os.path.abspath(input_dir)))
+        return True
```
rosswbrown: on success can you delete the unsigned taco file and move signed_connector.taco to connector.taco

packaging/package/jar_jdk_signer.py
```diff
@@ -0,0 +1,66 @@
...
+        return False
+
+    if not os.path.isfile(input_dir/taco_name):
+        logger.error("Signing Error: Taco file to be signed has been deleted or doesn't exist")
+        return False
+
+    return True
+
+
+def jdk_sign_jar(input_dir, taco_name, alias, keystore):
```
rosswbrown: Can you add some test coverage for this or open a follow-up item for next sprint.

packaging/package/jar_jdk_signer.py
```diff
@@ -0,0 +1,66 @@
...
+    :type alias: str
+
+    :param keystore: keystore path
+    :type keystore: str
+
+    :return: Boolean
+    """
+
+    if not check_jdk_environ_variable(JARSIGNER_EXECUTABLE_NAME):
+        logger.error("Error: jdk_create_jar: no jdk set up in PATH environment variable, "
```
lriggs: Since you have this error message in two places, maybe make it shared or possibly put into the check_ function.

packaging/package/jar_jdk_packager.py
```diff
@@ -5,24 +5,13 @@
 
 from pathlib import Path
 from .connector_file import ConnectorFile
+from .helper import check_jdk_environ_variable
 
 JAR_EXECUTABLE_NAME = "jar.exe"
 PATH_ENVIRON = "PATH"
```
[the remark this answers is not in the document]
jsun28: done

packaging/package/jar_jdk_signer.py
jsun28: done

packaging/package/jar_jdk_signer.py
jsun28: done

packaging/package/jar_jdk_signer.py
jsun28: Should be good with the latest commit, it will prompt the user to input before starting jarsigner, and those pwds will be passed down to jarsigner as input, and jarsigner's output will be logged.

packaging/package/jar_jdk_signer.py
```diff
@@ -28,6 +31,21 @@ def validate_signing_input(input_dir, taco_name, alias, keystore):
...
+def get_user_pwd(alias):
+    """
+    Let user input password for keystore and alias, which will be used as jarsigner subprocess input
+    """
+
+    ks_pwd = getpass.getpass(prompt='Enter keystore password: ', stream=None)
+    alias_pwd = getpass.getpass(prompt='Enter password for alias ' + alias + ":(RETURN if same as keystore password)",
+                                stream=None)
+
+    if alias_pwd == "" or alias_pwd == ks_pwd:
```
lriggs: What is this section doing? I don't see why line 44 and 46 are different.

jsun28: this section is asking user's password for alias and Keystore before spawning jarsigner, if the user's alias password is same with keystore pasword, the two elements in returning tuple are same, otherwise, first one is keystore pwd, the second one is alias pwd.

PR #182, opened 2019-07-02
[Docs] update sample ini to include [ConnectionTest]

lukewrites:
This PR is part of #174

PR #207, opened 2019-07-23
Add --verify flag to run smoke tests.

lukewrites:
Brings smoke tests into master. After merging can also merge #182 which updates docs to talk about the `ConnectionTest` section/smoke tests.


lukewrites:
Will merge into master as a "Squash and merge" so it only shows up as one commit rather than 78 of them.

PR #208, opened 2019-07-23
Added note about Null metadata.

lriggs:


docs/tdvt.md
```diff
@@ -241,7 +241,12 @@ ExpressionExclusions_Calcs =
 
 [StandardTests]
 
-[StaplesDataTest]
+[ConnectionTests]
```
pvanderknyff: Looks like you got some of Luke's commits in this one, are these ready to merge @lukewrites ?

lukewrites: Yup. Kathy's out sick and the smoke tests are in master. We can always edit it later if we want.

PR #213, opened 2019-07-26
add signing changes to readme

jsun28:
- fix --validate_only to --validate-only
- add signing to readme

### DIFF
--- a/packaging/README.md
+++ b/packaging/README.md
@@ -16,7 +16,7 @@ To package the connector and sign it:
 (.venv) PS connector-plugin-sdk\packaging> python -m package.package [path_to_folder] -a [alias_name] -ks [keystore_file_path]
 ```
 
-To package the connector only:
+To package the connector without signing:
 ```
 (.venv) PS connector-plugin-sdk\packaging> python -m package.package [path_to_folder] --package-only
 ```
--- a/packaging/README.md
+++ b/packaging/README.md
@@ -11,9 +11,14 @@ The recommendation is to install package within a Python virtual environment. Se
 
 ### Run package Module
 
-To package the connector:
+To package the connector and sign it:
 ```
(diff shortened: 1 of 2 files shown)


packaging/README.md
```diff
@@ -11,9 +11,14 @@ The recommendation is to install package within a Python virtual environment. Se
...
 ### Run package Module
 
-To package the connector:
+To package the connector and sign it:
 ```
-(.venv) PS connector-plugin-sdk\packaging> python -m package.package [path_to_folder]
+(.venv) PS connector-plugin-sdk\packaging> python -m package.package [path_to_folder] -a [alias_name] -ks [keystore_file_path]
+```
+
+To package the connector only:
```
pvanderknyff: Change to "To package the connector without signing" to be a bit more explicit in what it means. Right now someone might assume that without the flag we'd package something in addition to the connector.

packaging/package/package.py
```diff
@@ -16,11 +16,11 @@ def create_arg_parser():
     parser.add_argument('input_dir', help='path to directory of connector files to package and sign')
     parser.add_argument('-v', '--verbose', dest='verbose', action='store_true', help='verbose output', required=False)
     parser.add_argument('-l', '--log', dest='log_path', help='path of logging output', default='packaging_log.txt')
-    parser.add_argument('--validate_only', dest='validate_only', action='store_true', help='runs package validation steps only', required=False)
+    parser.add_argument('--validate-only', dest='validate_only', action='store_true', help='runs package validation steps only', required=False)
```
lukewrites: I think this change is already in 625b1bc which was committed directly to the `packaging` branch.

if you do `git pull origin packaging` it should show up in your branch. It's fine that you have it here, won't cause a merge conflict.

packaging/README.md
jsun28: Done, thanks.

PR #218, opened 2019-07-31
TFSID952474: Disable Aggregate:MoveGroupBysToMeasures since FuncallRewrite:RewriteConstantFuncall is also disabled

LingyanYin:
We have completed a new feature planned to turn on for 19.4. There is a rewrite rule added which has to work with RewriteConstantFuncall which is disabled in TDVT tests, so we also need to disable this one since it would cause some SQL text unable to be executed in DB2.

### DIFF
--- a/tdvt/tdvt/tabquery.py
+++ b/tdvt/tdvt/tabquery.py
@@ -3,6 +3,7 @@ import sys
 
 from .resources import *
 tab_cli_exe = ''
+logical_rewrite_appoption = ''
 
 def configure_tabquery_path():
     """Setup the tabquery path from ini settings."""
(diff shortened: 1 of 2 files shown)


LingyanYin:
> Thanks for the PR Lingyan, let's wait for a review from Logan but if he gives the thumbs up on the change it'd be helpful if you could fix the unit tests it looks like this change would break. You can see them in Travis CI: https://travis-ci.org/tableau/connector-plugin-sdk/jobs/566088933

Yeah. I took a look and it seems some changes in command line don't match the expected since I added another disabled rewrite in the command. I will check to see how to update those expected ones.

lukewrites:
> Yeah. I took a look and it seems some changes in command line don't match the expected since I added another disabled rewrite in the command. I will check to see how to update those expected ones.

I _think_ it'll be a quick fix since those tests are just string comparisons, but if they get weird ping me and I'm happy to help.

LingyanYin:
close it and discard it since Logan suggested another way of doing this

tdvt/tdvt/tabquery.py
```diff
@@ -3,6 +3,7 @@
 
 from .resources import *
 tab_cli_exe = ''
+logical_rewrite_appoption = ''
```
lriggs: I think it would be simpler if you took this out, then you wouldn't need the else clause on line 81 either.

tdvt/test/tdvt_test.py
```diff
@@ -253,9 +253,9 @@ def test_command_line_full(self):
         cmd_line = build_tabquery_command_line_local(work)
         cmd_line_str = ' '.join(cmd_line)
         if sys.platform in ('win32', 'cygwin'):
-            expected = 'tabquerytool.exe --expression-file-list my/output/dir\mytest\\tests.txt -d mytds.tds --combined --output-dir my/output/dir -DLogDir=my/output/dir\mytest -DOverride=ProtocolServerNewLog -DLogLevel=Debug -DLogicalQueryRewriteDisable=Funcall:RewriteConstantFuncall'  # noqa: E501
+            expected = 'tabquerytool.exe --expression-file-list my/output/dir\mytest\\tests.txt -d mytds.tds --combined --output-dir my/output/dir -DLogDir=my/output/dir\mytest -DOverride=ProtocolServerNewLog -DLogLevel=Debug ' + logical_rewrite_appoption  # noqa: E501
```
lriggs: Can't you just leave the test file like it is (ie revert these changes)?

PR #220, opened 2019-07-31
TFSID951139: [TDVT] Update sample plugins in SDK

saugatt:
Update connector_plugin_manifest_latest.xsd 
Add min-version-tableau in sample plugins
Add vendor-information in manifest.xml for sample plugins

### DIFF
--- a/validation/connector_plugin_manifest_latest.xsd
+++ b/validation/connector_plugin_manifest_latest.xsd
@@ -94,9 +94,40 @@
       <xs:pattern value="(\d+)((\.)(\d+))*((-)(\d+))*"/>
     </xs:restriction>
   </xs:simpleType>
+  <xs:group name="VendorInformation-G">
+    <xs:sequence>
+      <xs:element minOccurs="0" name="vendor-information">
+        <xs:complexType>
+          <xs:sequence>
+            <xs:group ref="Company-G"/>
+            <xs:group ref="SupportLink-G"/>
+          </xs:sequence>
+        </xs:complexType>
+      </xs:element>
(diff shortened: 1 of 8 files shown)


jsun28:
I think we're going to merge this change to the dev branch, instead of packaging branch. isn't it?

validation/connector_plugin_manifest_latest.xsd
```diff
@@ -94,9 +94,40 @@
       <xs:pattern value="(\d+)((\.)(\d+))*((-)(\d+))*"/>
     </xs:restriction>
   </xs:simpleType>
+  <xs:group name="VendorInformation-G">
```
jsun28: For this VendorInformation-G group, we have a separate xsd file in monolith, https://opengrok/source/xref/teams_near/modules/XSD/VendorInfo.xsd. we probably want to keep it consistent with the monolith code.

saugatt: The update is just the copy from an updated generated file in near\tableau-1.3\res\FileFormat\connector_plugin_manifest_latest.xsd without the comments.

saugatt: Since I branched off of packaging, merging to dev will merge everything from packaging that has not yet been merged to dev.

saugatt:
> Double check that the packager will package the samples properly before merging.

lukewrites:
@jsun28 I think it’s ok to merge it in here. If we want it in dev before the packaging tools is merged we can cherry pick the commits in.

PR #221, opened 2019-07-31
Update dev branch's xsd to be same as monolith

jsun28:
Update xsd to be same as monolith

### DIFF
--- a/validation/connector_plugin_manifest_latest.xsd
+++ b/validation/connector_plugin_manifest_latest.xsd
@@ -101,6 +101,7 @@
           <xs:sequence>
             <xs:group ref="Company-G"/>
             <xs:group ref="SupportLink-G"/>
+            <xs:group ref="DriverDownloadLink-G" minOccurs="0"/>
           </xs:sequence>
         </xs:complexType>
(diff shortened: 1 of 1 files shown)


validation/connector_plugin_manifest_latest.xsd
```diff
@@ -94,9 +94,40 @@
...
+        <xs:complexType>
+          <xs:attribute name="url" type="xs:anyURI" use="required"/>
+        </xs:complexType>
+      </xs:element>
+    </xs:sequence>
+  </xs:group>
   <xs:element name="connector-plugin">
     <xs:complexType>
       <xs:sequence>
+        <xs:group ref="VendorInformation-G"/>
```
rosswbrown: So we're making this element required in the SDK?  The postgres_odbc/jdbc samples will need to be updated also if so.  How is this made optional on the platform side?

jsun28: No, it is still optional here. See line 99 `<xs:element minOccurs="0" name="vendor-information">`. 
XSD after changes should be same as monolith's xsd.

PR #222, opened 2019-08-01
Packaging tool improvements before shipping

lukewrites:
This PR does the following:
* Change the `--log` arg. Previously it took a path to a specific file. Now, it does what the label says, and accepts a path to where the file is saved.
* Raises a warning at the command line and in the logs when `--validate-only` and `--package-only` are run together.
* Minor changes to README
* Brings in pep8speaks so it'll stop nagging us about 120 char lines

There are whitespace changes in this PR; you can have GitHub ignore them by selecting the following:
Snag_b3a764

### DIFF
--- a/packaging/package/package.py
+++ b/packaging/package/package.py
@@ -34,7 +34,7 @@ def create_arg_parser():
 def main():
     parser = create_arg_parser()
     args = parser.parse_args()
-    log_file = args.log_path + 'packaging_log.txt'
+    log_file = args.log_path + '/packaging_log.txt'
     logger = init_logging(log_file, args.verbose)
 
     path_from_args = Path(args.input_dir)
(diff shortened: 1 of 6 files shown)


rosswbrown:
Can you remove the Notes section at the bottom of the README as well.

packaging/package/package.py
```diff
@@ -12,44 +13,53 @@
 
 
 def create_arg_parser():
-    parser = argparse.ArgumentParser(description="Tableau Connector Packaging Tool: package and sign connector files into a single Tableau Connector (" + PACKAGED_EXTENSION + ") file.")
+    parser = argparse.ArgumentParser(description="Tableau Connector Packaging Tool: package and sign connector files into a single Tableau Connector (" + PACKAGED_EXTENSION + ") file.")  # noqa: E501
```
pvanderknyff: What is #noqa: e501 mean? Is it telling the PEP bot to ignore this line or something?

packaging/package/xml_parser.py
```diff
@@ -38,20 +38,20 @@ def generate_file_list(self):
             list[ConnectorFile] -- list of files to package
             -- Returns none if any of the files are invalid, or the files do not agree on the name
         """
-        
-        logging.debug("Generating list of files to package...")
-        
+
+        logging.debug("Generating list of for validation and/or packaging...")
```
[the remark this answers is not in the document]
lukewrites: whoops meant for it to read

> list **of** files

packaging/package/package.py
lukewrites: yup. you can do a plain `# noqa`, including the pep8 error code is optional but makes sure it ignores what you're intending.

PR #226, opened 2019-08-07
Stamp min support tableau version while packaging

jsun28:
Stamp min support tableau version while packaging(updated)
- Do the stamping after packaging
- Make a copy of the original manifest file, after packaging, recover it to keep original manifest unchanged 

Test it: 
py -3 -m connector_packager.package tests\test_resources\valid_connector --package-only

### DIFF
--- a/connector-packager/connector_packager/jar_jdk_packager.py
+++ b/connector-packager/connector_packager/jar_jdk_packager.py
@@ -2,13 +2,74 @@ import os
 import logging
 import subprocess
 import shutil
+import xml.etree.ElementTree as ET
 
 from pathlib import Path
 from .connector_file import ConnectorFile
 from .helper import check_jdk_environ_variable
+from .version import __min_version_tableau__
 
 JAR_EXECUTABLE_NAME = "jar.exe"
 logger = logging.getLogger(__name__)
+MANIFEST_FILE_TYPE = "manifest"
+MANIFEST_FILE_NAME = "manifest.xml"
+MANIFEST_FILE_COPY_NAME = "manifest_copy.xml"
+MANIFEST_ROOT_ELEM = "connector-plugin"
+MIN_TABLEAU_VERSION_ATTR = "min-version-tableau"
+
+
+def stamp_min_support_version(input_dir, file_list, jar_filename):
+    """
+    Stamp of minimum support version to the connector manifest in packaged jar file
+
+    :param input_dir: source dir of files to be packaged
+    :type input_dir: Path
+
+    :param file_list: files need to be packaged
+    :type file_list: list of ConnectorFile
+
+    :param jar_filename: filename of the created JAR
+    :type jar_filename: str
+
+    :return: Boolean
+    """
+
+    # find manifest in connector file list
+    manifest_file = None
+    for file in file_list:
+        if file.file_type == MANIFEST_FILE_TYPE and file.file_name == MANIFEST_FILE_NAME:
+            manifest_file = file
+            break
+    if not manifest_file:
+        logger.info("Can not find manifest.xml in input directory while packaging")
+        return False
+
+    # make a copy of manifest file
+    shutil.copyfile(input_dir / manifest_file.file_name, input_dir / MANIFEST_FILE_COPY_NAME)
+
+    # stamp the original manifest file
+    manifest = ET.parse(input_dir / manifest_file.file_name)
+    plugin_elem = manifest.getroot()
+    if plugin_elem.tag != MANIFEST_ROOT_ELEM:
+        logger.info("Manifest's root element has been modified after xml validation")
+        return False
+    plugin_elem.set(MIN_TABLEAU_VERSION_ATTR, __min_version_tableau__)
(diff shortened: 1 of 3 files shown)


packaging/package/jar_jdk_packager.py
```diff
@@ -2,15 +2,63 @@
...
+    :type file_list: list of ConnectorFile
+
+    :return: Boolean
+    """
+
+    # find manifest in connector file list
+    manifest_file = None
+    for file in file_list:
+        if file.file_type == MANIFEST_FILE_TYPE and file.file_name == MANIFEST_FILE_NAME:
+            manifest_file = file
```
lukewrites: should we add a `break` here since there should only be one `ConnectorFile` obj that corresponds to a `manifest_file`?
```suggestion
            manifest_file = file
            break
```

packaging/package/jar_jdk_packager.py
```diff
@@ -2,15 +2,63 @@
...
+    if not manifest_file:
+        logging.info(
+            "Can not find manifest file in input directory, packaging failed")
+        return False
+
+    # make a copy of manifest file
+    shutil.copyfile(str(input_dir) + "\\" + manifest_file.file_name,
+                    str(input_dir) + "\\" + MANIFEST_FILE_COPY_NAME)
```
lukewrites: I'm not sure if hardcoding `\` as the delimiter is the right choice here since that's a Windows thing. Will it cause problems on mac/linux, which use `/`?

Maybe we could break out the source & dest dirs as vars made of Path objects & then put their strings here?

packaging/package/jar_jdk_packager.py
```diff
@@ -2,15 +2,63 @@
...
+
+    manifest = ET.parse(str(input_dir) + "\\" + manifest_file.file_name)
```
lukewrites: same as above

packaging/package/jar_jdk_packager.py
```diff
@@ -2,15 +2,63 @@
...
+    plugin_elem = manifest.getroot()
+    if plugin_elem.tag != MANIFEST_ROOT_ELEM:
+        logging.info(
+            "Manifest's root element has been modified after xml validation")
+        return False
+
+    plugin_elem.set(MIN_TABLEAU_VERSION_ATTR, __min_version_tableau__)
+    manifest.write(str(input_dir) + "\\" + manifest_file.file_name,
```
lukewrites: ☝️

packaging/package/jar_jdk_packager.py
```diff
@@ -34,12 +82,17 @@ def jdk_create_jar(source_dir, files, jar_filename, dest_dir):
         return False
 
     abs_source_path = os.path.abspath(source_dir)
-    logging.debug("Start packaging " + jar_filename + " from " + str(abs_source_path) + " using JDK")
+    logging.debug("Start packaging " + jar_filename +
+                  " from " + str(abs_source_path) + " using JDK")
```
lukewrites: this is less than 120 chars, should be fine on one line

packaging/package/jar_jdk_packager.py
pvanderknyff: Use the python Path object:
path_from_args = Path(args.input_dir)

Which lets you do things like this:
path_to_file = folder_path / file_to_test.file_name

This is how we do it in other parts of the packager, and is a platform-agnostic way of dealing with file paths.

jsun28: Thanks!

packaging/package/jar_jdk_packager.py
jsun28: Thanks, done.

packaging/package/jar_jdk_packager.py
jsun28: Fixed with Patrick's suggestion

lukewrites:
you're going to need to pull in `packaging` to resolve the merge conflicts. shouldn't be anything major.

jsun28:
@lukewrites sure, no worries.
@rosswbrown I updated the commits, now the stamping is happening after packaging, like we discussed this morning. let me know if you see anything wrong.
Thanks!

lukewrites:
If the merge conflicts are too out of hand from my renaming PR, I can back it out, let this one in, then merge back. Just lmk

jsun28:
It is all good. I have already pulled your changes and rebased, so my current PR should have no conflict with latest packaging branch.

connector-packager/connector_packager/jar_jdk_packager.py
```diff
@@ -2,13 +2,74 @@
...
+import xml.etree.ElementTree as ET
 
 from pathlib import Path
 from .connector_file import ConnectorFile
 from .helper import check_jdk_environ_variable
+from .version import __min_version_tableau__
 
 JAR_EXECUTABLE_NAME = "jar.exe"
 logger = logging.getLogger(__name__)
+MANIFEST_FILE_TYPE = "manifest"
```
rosswbrown: reuse MANIFEST_FILE_TYPE for defining MANIFEST_FILE_NAME and MANIFEST_FILE_COPY_NAME

PR #227, opened 2019-08-07
Rename packaging

lukewrites:
Change name from `packaging` to `connector-packager`. The python library itself is `connector_packager`, since Python doesn't like using a `-` in library names.

Also updated README.


connector-packager/tests/test_jar_signer.py
```diff
@@ -0,0 +1,38 @@
+import unittest
```
jsun28: Is this test_jar_signer.py already in the packaging branch?

lukewrites: looks like it isn't did I pull in something new?

rosswbrown: Ensure you've got 4549b1a

lukewrites: My mistake, it's in there. I think it's showing up as "new" in this PR because it's been moved.

lukewrites:
Figured it out; git thinks the original test file was deleted because there number of lines are different.

PR #232, opened 2019-08-09
Add type hints to `connector-packager`

lukewrites:
This PR adds type hints to all function & method defs in the `connector-packager` library. The addition of these type hints will allow type errors to be raised during development, if your editor is set up correctly. If you use VS Code, [Pyright](https://marketplace.visualstudio.com/items?itemName=ms-pyright.pyright) will allow you to take advantage of type hints. PyCharm should do so automatically.

Example definition & raised type errors:
image

image

Tests are passing on the PR as of 72b0d60 but we should poke at it to make sure everything's OK.


connector-packager/connector_packager/helper.py
```diff
@@ -1,12 +1,17 @@
 import os
 import logging
+
 from pathlib import Path
+from typing import Optional
```
lukewrites: `Optional` accepts either the value(s) specified or `None`

connector-packager/connector_packager/connector_file.py
```diff
@@ -12,8 +12,8 @@ def extension(self) -> str:
         # Remove the leading period.
         return ext if not ext else ext[1:]
 
-    def __lt__(self, other):
+    def __lt__(self, other: 'ConnectorFile') -> bool:
```
lukewrites: this is a [forward reference](https://www.python.org/dev/peps/pep-0484/#forward-references)

connector-packager/connector_packager/helper.py
```diff
@@ -16,11 +21,11 @@ def check_jdk_environ_variable(exe_name):
             return True
 
     logger.error("Java Error: jdk_create_jar: no jdk set up in PATH environment variable, "
-                     "please download JAVA JDK and add it to PATH")
+                 "please download JAVA JDK and add it to PATH")
     return False
 
 
-def init_logging(log_path, verbose):
+def init_logging(log_path: str, verbose: Optional[bool]) -> logging.Logger:
```
chrisjunlee: Why "**verbose: Optional[bool]**" vs "**verbose: bool = False**"?

connector-packager/connector_packager/jar_jdk_packager.py
```diff
@@ -33,7 +35,7 @@ def jdk_create_jar(source_dir, files, jar_filename, dest_dir):
     if not check_jdk_environ_variable(JAR_EXECUTABLE_NAME):
         return False
 
-    abs_source_path = os.path.abspath(source_dir)
+    abs_source_path = source_dir.resolve()
```
chrisjunlee: I like this, looks cleaner. Does it have the same functionality as "os.path.abspath(source_dir)", and is it the pythonic way forward?

connector-packager/connector_packager/jar_jdk_packager.py
```diff
@@ -46,8 +48,7 @@ def jdk_create_jar(source_dir, files, jar_filename, dest_dir):
     p = subprocess.Popen(args, cwd=abs_source_path)
     p.wait()
 
-    shutil.move(abs_source_path/Path(jar_filename), dest_dir/jar_filename)
+    shutil.move(abs_source_path/jar_filename, dest_dir/jar_filename)
```
chrisjunlee: dest_dir is the absolute path right?

connector-packager/connector_packager/jar_jdk_signer.py
```diff
@@ -34,7 +38,7 @@ def validate_signing_input(input_dir, taco_name, alias, keystore):
     return True
 
 
-def get_user_pwd(alias):
+def get_user_pwd(alias: str) -> Tuple[bytes, bytes]:
```
chrisjunlee: What is **Tuple[bytes, bytes]**? List of (byte, byte) tuples?

connector-packager/connector_packager/jar_jdk_signer.py
```diff
@@ -49,12 +53,12 @@ def get_user_pwd(alias):
         return str.encode(ks_pwd + "\n"), str.encode(alias_pwd + "\n")
 
 
-def jdk_sign_jar(input_dir, taco_name, alias, keystore):
+def jdk_sign_jar(input_dir: Path, taco_name: str, alias: str, keystore: str) -> bool:
```
chrisjunlee: I would expect the hinting to be the same as validate_signing_input (alias and keystore being _Optional[str]_). Is the difference intentional?

connector-packager/tests/jar_packager.py
```diff
@@ -33,11 +34,11 @@ def create_jar(source_dir, files, jar_filename, dest_dir):
...
 
     # if dest dir doesn't exist, then create it
-    if not os.path.exists(dest_dir):
-        os.makedirs(dest_dir)
+    if not dest_dir.exists():
+        dest_dir.mkdir()
         logging.debug("Creating destination directory " + str(dest_dir))
 
-    with ZipFile(os.path.join(dest_dir, Path(jar_filename)), "w") as jar:
+    with ZipFile(dest_dir / jar_filename, "w") as jar:
```
chrisjunlee: I'm a bit concerned about the implicit '/' concat for file path since windows uses '\'. I believe os.path.join concats using the appropriate delimiter.

connector-packager/connector_packager/jar_jdk_packager.py
lukewrites: It's a `Path` method that does the same thing as `os.path.abspath`! 😁

connector-packager/connector_packager/jar_jdk_packager.py
lukewrites: Yeah, it's a `Path` object per the fn def

connector-packager/connector_packager/jar_jdk_signer.py
lukewrites: Python is doing something unpython here, I think. It means it's returning a `(bytes, bytes)` tuple. A list of tuples would be `List[Tuple[bytes, bytes]]`. Yucky, right?

connector-packager/tests/jar_packager.py
lukewrites: The slash operator [works differently/cleverly with `Path` objects](https://docs.python.org/3/library/pathlib.html#operators).

connector-packager/connector_packager/helper.py
lukewrites: That's probably better, I _think_ that `None` is being passed by default sometimes? I'll dbl check in the IDE.

connector-packager/connector_packager/jar_jdk_signer.py
chrisjunlee: Yucky, but oddly pythonic in the sense that I could read it

connector-packager/connector_packager/jar_jdk_signer.py
lukewrites: Let's dbl check with @jsun28 but I _think_ they're all required in the function. Will change if not :)

connector-packager/connector_packager/helper.py
lukewrites: fixed in f04d655

connector-packager/connector_packager/jar_jdk_signer.py
jsun28: All parameters in this function are required.

lukewrites:
I rerequested review since (a) there have been some big changes and (b) this was a pretty stale PR

PR #234, opened 2019-08-12
Initial Authentication Modes doc

rosswbrown:
Document the supported auth modes with sample connection dialog, connection resolver.  Also include notes about the hive specific constraints.

### DIFF
--- a/docs/auth-modes.md
+++ b/docs/auth-modes.md
@@ -1,0 +1,263 @@
+---
+title: Authentication Modes
+---
+
+The authentication mode chosen below influences how and when a user is prompted to enter data source credentials. The primary scenarios where authentication occurs:
+
+- Creating a new connection with the connection dialog
+- Opening a workbook and reconnecting to the data source
+- Publishing a workbook or data source to Tableau Server
+
+A combination of Connection Dialog and Connection Resolver changes implement each of these modes.  
+
+## Supported Authentication Modes
+
+### No Authentication
+User is never prompted for credentials.
+
+```xml
+<!-- Connection Dialog -->
+<connection-dialog class='sample'>
+    <connection-config>
+        <authentication-mode value='None' />
+        <authentication-options>
+            <option name="None" default="true" value="auth-none" />
+        </authentication-options>
+        ...
+    </connection-config>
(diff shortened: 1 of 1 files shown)


docs/auth-modes.md
```diff
@@ -0,0 +1,263 @@
...
+    </connection-resolver>
+    ...
+</tdr>
+```
+
+### Multiple Authentication Modes
+
+User is prompted for which authentication option to use, then a set of fields appear, conditional on that option.  Depending on the option selected the user may or may not be prompted for credentials when reconnecting to the data source.
+
+Support authention options are:
```
kpwilson: authention = authentication

docs/auth-modes.md
```diff
@@ -0,0 +1,263 @@
...
+            <option name="None" value="auth-none" />
+            <option name="Username" value="username" />
+            <option name="UsernameAndPassword" value="username-password" default="true" />
+        </authentication-options>
+        ...
+    </connection-config>
+</connection-dialog>
+```
+
+Note: the ```value``` attribute value for all options are customizable by connector author, except None, which is required to be ```auth-none```.  These option values are the persisted value of the authentication attribute in a Tableau Workbook (twb) or Tableau Data Source (tds) file.
```
kpwilson: are -> is  "...attribute value for all options is customizable by ..."  I'd also suggest that you lower-case "workbook" and "data source"

docs/auth-modes.md
```diff
@@ -0,0 +1,263 @@
...
+
+    ...
+})
+```
+
+## Considerations for 'hadoophive' and 'spark' base classes
+
+When using 'hadoophive' or 'spark' as a base class there are additional constraints and requirements in the Connection Dialog and Connection Resolver.
+
+The Connection Dialog authentication options ```option``` element's ```value``` attribute's value must match below.
```
kpwilson: I'm having a hard time parsing this sentence -- is it possible to construct the sentence to say "this" must match "this"?

rosswbrown: I was having trouble with this section as well.  I think its a little better now.

PR #235, opened 2019-08-13
2019.2 holding

lukewrites:


samples/plugins/postgres_jdbc/manifest.xml
```diff
@@ -1,6 +1,10 @@
 <?xml version='1.0' encoding='utf-8' ?>
 
 <connector-plugin class='postgres_jdbc' superclass='jdbc' plugin-version='0.0.0' name='PostgreSQL JDBC' version='18.1'>
+  <vendor-information>
+      <company name="Company Name"/>
+      <support-link url = "support-link-url"/>
```
pvanderknyff: We should use http://example.com for both connectors so that they can click on the link and not get an error.

rosswbrown:
Can you confirm 2019.2 will load these samples.

samples/plugins/postgres_odbc/manifest.xml
```diff
@@ -1,6 +1,10 @@
 <?xml version='1.0' encoding='utf-8' ?>
 
 <connector-plugin class='postgres_odbc' superclass='odbc' plugin-version='0.0.0' name='PostgreSQL ODBC' version='18.1'>
+  <vendor-information>
+      <company name="Company Name"/>
+      <support-link url = "support-link-url"/>
```
rosswbrown: "http://example.com"

PR #241, opened 2019-08-14
document localization support

sxlin:
Added a doc page for localization support.

### DIFF
--- a/docs/localize.md
+++ b/docs/localize.md
@@ -1,0 +1,58 @@
+---
+title: Localize Your Connector
+---
+
+Tableau products have been localized to English (United States), English (United Kingdom), French, German, Brazilian Portuguese, Spanish, Korean, Japanese, simplified Chinese and traditional Chinese.
+
+Connector SDK supports localization of connectors to the languages mentioned above.
+
+The strings showing up on Tableau UI can be localized. For example, the connector name and the prompts in connection dialog.
+## Specify localized Strings
+Localized strings are specified using tag `@string/<string_id>/`, for example, `@string/database_prompt/`. You can use the tag anywhere a string is defined in `manifest.xml` or Tableau Custom Dialog (.tcd) file.
+
+Here is an example of a Tableau Custom Dialog (.tcd) file.
+```xml
+<!-- Connection Dialog -->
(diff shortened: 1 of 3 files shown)


docs/localize.md
```diff
@@ -0,0 +1,58 @@
...
+The resource files must be encoded in UTF-8.
+
+If the resource file for a language is missing, or a string translation is missing in a resource file if you have any localized strings in your connector, Tableau will fall back to use translation in English (United States).  You must have resources-en-US.xml included in your connector, that file must include all localized strings.
+ 
+Here is an example of resource file `resources-es_ES.xml`.
+ 
+ ```xml
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <string name="port_prompt">Puerto:</string>
```
pvanderknyff: Can we also have the example in English as well? (For example, here is the en_US file, and the corresponding es_ES file). Just to explicitly show a translation.

Not completely necessary, but a screenshot of English vs. Spanish connection dialog might be good too, to show visually what these files change.

lriggs: I like that idea too. It would be a minimal example that they can easily build on to other languages.

docs/localize.md
```diff
@@ -0,0 +1,69 @@
...
+For the best user experience with Tableau, it is highly recommended that translation support is included.
+## Specify localized Strings
+Localized strings are specified using tag `@string/<string_id>/`, for example, `@string/database_prompt/`. You can use the tag anywhere a string is defined in `manifest.xml` or Tableau Custom Dialog (.tcd) file.
+
+Here is an example of a Tableau Custom Dialog (.tcd) file with localized strings.
+```xml
+<!-- Connection Dialog -->
+<connection-dialog class='sample'>
+  <connection-config>
+    <authentication-mode value='Basic' />
```
rosswbrown: To keep the sample snippets short and simple I've been using ..., I think you could replace all the auth stuff with ... here.

kpwilson:
I would like to review, but doubt that I can get to it this week. If you need to publish sooner rather than later, please do. I can follow up with questions and suggestions after the fact.

PR #256, opened 2019-08-21
Update xsd's vendor info section as required

jsun28:


samples/plugins/mysql_odbc/manifest.xml
```diff
@@ -4,6 +4,7 @@
   <vendor-information>
     <company name='TestCompany' />
     <support-link url='http://example.com' />
+    <driver-download-link url="http://driverexamples.com"/>
```
rosswbrown: Use `http://drivers.example.com`.  It will 404, but I think thats ok and I'd rather not use random domains in the samples.

jsun28: yeah, makes sense. Fixed.

PR #257, opened 2019-08-22
Update vendor info stuff in packaging branch

jsun28:
1. update xsd to date
2. update test resource manifest files
Let me know if I missed anyting.


validation/connector_plugin_manifest_latest.xsd
```diff
@@ -96,11 +98,12 @@
   </xs:simpleType>
   <xs:group name="VendorInformation-G">
     <xs:sequence>
-      <xs:element minOccurs="0" name="vendor-information">
+      <xs:element name="vendor-information">
```
rosswbrown: Let's hold off on this file's change.  I'd rather you merge your pull request into dev, then we can update the packaging branch with the latest changes from dev to pick this up

jsun28: Sure, that makes sense. Fixed.

PR #266, opened 2019-09-03
Update postgres_odbc connectionBuilder.js

jainam1995:
Add code snippet to use odbc-connect-string-extras set in postgres_odbc

### DIFF
--- a/samples/plugins/postgres_odbc/connectionBuilder.js
+++ b/samples/plugins/postgres_odbc/connectionBuilder.js
@@ -13,10 +13,10 @@
     params["Fetch"] = "2048";
 
     var odbcConnectStringExtrasMap = {};
-    const attributeODBCConnectionStringExtras = "odbc-connect-string-extras";
-    if (attributeODBCConnectionStringExtras in attr) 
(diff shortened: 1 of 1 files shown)


samples/plugins/postgres_odbc/connectionBuilder.js
```diff
@@ -12,6 +12,16 @@
     params["UseDeclareFetch"] = "1";
     params["Fetch"] = "2048";
 
+    var odbcConnectStringExtrasMap = {};
```
rosswbrown: Can you define the string as a const?

`const attributeODBCConnectionStringExtras = "odbc-connect-string-extras";`

jainam1995: Fixed

samples/plugins/postgres_odbc/connectionBuilder.js
```diff
@@ -12,6 +12,17 @@
...
+    const attributeODBCConnectionStringExtras = "odbc-connect-string-extras";
```
rosswbrown: rename attributeODBCConnectStringExtras, missed this in my comment earlier.

PR #268, opened 2019-09-04
Remove failed tests if they succeed on retry

pvanderknyff:
If the test fails in teamcity and succeeds on retry, we will add the test to the successful tests, but not remove it from the failed test list in tdvt_ouput_combined.json. I think this is why teamcity counts the test cases that succeed on retry as failed. This commit will strip out the failed test if it is in the newly succeeding tests.


lukewrites:
You're right, a list comp would probably be unwieldy in this case.

PR #274, opened 2019-09-10
added CONTAINS test for empty substring

jd-dean:
added a test for CONTAINS with an empty substring

### DIFF
--- a/tdvt/tdvt/exprtests/standard/setup.string.contains.txt
+++ b/tdvt/tdvt/exprtests/standard/setup.string.contains.txt
@@ -1,2 +1,3 @@
 CONTAINS(str2, "e")
 CONTAINS(str1, "IND")
+CONTAINS(str2, LEFT(str2, 0))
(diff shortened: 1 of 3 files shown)


tdvt/tdvt/exprtests/standard/setup.string.contains.txt
```diff
@@ -1,2 +1,3 @@
 CONTAINS(str2, "e")
 CONTAINS(str1, "IND")
+CONTAINS(str2, LEFT(str2, 0))
```
lriggs: I recommend adding this as a new setup file, something like setup.string.contains.empty.txt. That way you could add two new expected files and leave the originals unchanged. This simplifies maintenance since you don't have to duplicate existing test case results in order to add a new expected file.

lukewrites:
@jd-dean how soon do you need this in the monolith? We might want to base this PR against `master` if you want it in soon.

jd-dean:
there's no rush on this.  i'm going to restructure this per Logan's suggestion.

jd-dean:
i've created a new test and three new expected files as you suggested.  will these updates automatically replace the files in the pull request?

PR #283, opened 2019-09-23
Add initial sql + impersonate attributes to samples

pvanderknyff:
For the 2019.4 partner-built plugins, this was something I had to add to all of them.

### DIFF
--- a/samples/plugins/mysql_odbc/connectionResolver.tdr
+++ b/samples/plugins/mysql_odbc/connectionResolver.tdr
@@ -6,6 +6,7 @@
     </connection-builder>
     <connection-normalizer>
       <required-attributes>
+        <setImpersonateAttributes/>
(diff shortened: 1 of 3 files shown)


samples/plugins/mysql_odbc/connectionResolver.tdr
```diff
@@ -6,6 +6,7 @@
     </connection-builder>
     <connection-normalizer>
       <required-attributes>
+        <setImpersonateAttributes/>
```
lriggs: Did we have any documentation about what this is or why someone would want to set it?

pvanderknyff: I have some doc changes in the pipe as a side thing for hackathon that include this scenario.

PR #284, opened 2019-09-23
Documentation tweaks

pvanderknyff:
- Removed outdated section on web authoring
- Changed headings of resolver methods since the normalizer might not be a .js file
- Change normalizer example to the recommended xml implementation
- Highlight that odbc and jdbc need a dialect file
- Change name of "Share your connector" page to "Run your connector". The most it says about sharing is "don't do it" since it's in beta, and actually running the connector is a bit hidden even with Logan's addition to the getting started page. This is the one I'm not as sure of, I'm open to suggestions on the best way to highlight running the connector (maybe split it into two separate pages?)

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -32,10 +32,6 @@ If your connection is very similar to an existing database such as PostgreSQL or
 
 You might find other super class values in the workbook XML from an existing Tableau connection, but we don't recommend using them; they haven't been tested.
 
-### Features based on superclass
-
-Web Authoring (creating a new connection from the web) is not currently available for all connector superclasses. If your connector is based on 'odbc' or 'jdbc,' then you can publish your workbook or datasource from desktop to server, but you can't create a new connection directly on server. Connectors based on 'mysql_odbc' do support web authoring because this ability is inherited from the mysql code they are based on. 
-
 ## Choosing a dialect
 
 The dialect determines what SQL is generated for various Tableau actions. Choosing the right dialect is a critical part of writing a connector.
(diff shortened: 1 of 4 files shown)


kpwilson:
These changes look good to me. We might make some tweaks as the additional content is added for how to sign and package connectors and signature verification.

docs/design.md
```diff
@@ -70,7 +66,8 @@ The dialect determines what SQL is generated for various Tableau actions. Choosi
 - VerticaDialect
 
 ### Should I create a dialect definition file?
-You don’t need a dialect definition file if your connector uses the same SQL dialect as the connector it’s based on, such as mysql_odbc. However, to make changes to an existing Tableau dialect or define a new dialect, you’ll need a new dialect definition file. For more information, see [Create a Tableau Dialect Definition (TDD) File]({{ site.baseurl }}/docs/dialect). 
+You don’t need a dialect definition file if your connector uses the same SQL dialect as the connector it’s based on, such as mysql_odbc. However, to make changes to an existing Tableau dialect or define a new dialect, you’ll need a new dialect definition file. For more information, see [Create a Tableau Dialect Definition (TDD) File]({{ site.baseurl }}/docs/dialect).
```
lriggs: This might be a little clearer if you started it with something like 'If you want to customize the generated SQL, or your connector inherits from 'odbc' or 'jdbc' then you need a dialect, otherwise you can use the parent dialect.'  That way the two exceptions are together. Just an idea.

pvanderknyff: Tweaked the wording here a bit.

PR #286, opened 2019-09-24
Add properties builder to postgres_jdbc

pvanderknyff:
This is the same pattern used for MaxCompute. Tested in 2019.3.

### DIFF
--- a/samples/plugins/postgres_jdbc/connectionResolver.tdr
+++ b/samples/plugins/postgres_jdbc/connectionResolver.tdr
@@ -17,5 +17,8 @@
         </attribute-list>
       </required-attributes>
     </connection-normalizer>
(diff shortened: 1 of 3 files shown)


samples/plugins/postgres_jdbc/connectionResolver.tdr
```diff
@@ -17,5 +17,8 @@
         </attribute-list>
       </required-attributes>
     </connection-normalizer>
+    <connection-properties>
+        <script file='connectionProperties.js'/>
```
rosswbrown: can you clean-up the whitespace in this file too.

pvanderknyff: Fixed, thanks!

PR #293, opened 2019-09-27
Fix ssl for postgres examples

pvanderknyff:
The ODBC sample had a checkbox that did nothing. Added the logic in the connection builder, and then added the checkbox and logic to the JDBC sample.

### DIFF
--- a/samples/plugins/postgres_odbc/connectionBuilder.js
+++ b/samples/plugins/postgres_odbc/connectionBuilder.js
@@ -12,6 +12,12 @@
     params["UseDeclareFetch"] = "1";
     params["Fetch"] = "2048";
 
+    if (attr[connectionHelper.attributeSSLMode] == "require")
+    {        
+        params["sslmode"] = "require";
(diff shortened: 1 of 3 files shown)


rosswbrown:
You'll need to update the required attributes as well.

samples/plugins/postgres_odbc/connectionBuilder.js
```diff
@@ -12,6 +12,12 @@
     params["UseDeclareFetch"] = "1";
     params["Fetch"] = "2048";
 
+    if (attr[connectionHelper.attributeSSLMode] == "require")
+    {        
+        params["sslmode"] = "require";
```
rosswbrown: required attribute: `<attr>sslmode</attr>`

pvanderknyff: Thanks for catching that, opened https://github.com/tableau/connector-plugin-sdk/pull/294

PR #307, opened 2019-10-08
Refactor command line arguments.

lriggs:


tdvt/tdvt/tdvt.py
```diff
@@ -400,41 +403,68 @@ def usage_text():
     ie cast_calcs.*.tds for cast_calcs.sqlserver.tds etc.
 
     Run one logical query test against many datasources
-        py -3 -m tdvt.tdvt --logp logicaltests/setup/calcs/setup.BUGS.B1713.?.xml --tdp cast_calcs.*.tds --run postgres
+        run-pattern --logp logicaltests/setup/calcs/setup.BUGS.B1713.?.xml --tdp cast_calcs.*.tds postgres
 
     This can be combined with * to run an arbitrary set of 'correct' logical query tests against a datasources
-        py -3 -m tdvt.tdvt --logp logicaltests/setup/calcs/setup.BUGS.*.?.xml --tdp cast_calcs.*.tds --run postgres
+        run-pattern --logp logicaltests/setup/calcs/setup.BUGS.*.?.xml --tdp cast_calcs.*.tds  postgres
```
lukewrites: Not sure if the two spaces between `cast_calcs.*.tds` and `postgres` are intentional, but if not: 
```suggestion
        run-pattern --logp logicaltests/setup/calcs/setup.BUGS.*.?.xml --tdp cast_calcs.*.tds postgres
```

tdvt/tdvt/tdvt.py
```diff
@@ -621,40 +662,40 @@ def run_file(run_file, output_dir, threads, args):
...
+    elif is_test(args):
+        if args.command == 'run' and args.run_file:
+            output_dir = os.getcwd()
+            max_threads = get_level_of_parallelization(args)
+            sys.exit(run_file(args.run_file, output_dir, max_threads, args))
+        error_code = run_desired_tests(args, ds_registry)
+        sys.exit(error_code)
+    
+    logging.error("Could not interpert arguments. Nothing done.")
+    sys.exit(-1)
```
lukewrites: Would it make sense to do something like this here?

```suggestion
    else:
        logging.error("Could not interpret arguments. Nothing done.")
        p = create_parser()
        p.print_help()
        sys.exit(-1)
```
That would print the help text if the args are missing/weird, which might save the user a few keystrokes.

lriggs: Done

PR #312, opened 2019-10-10
Add run-file command to more cleanly seperate this.

lriggs:


tdvt/tdvt/tdvt.py
```diff
@@ -492,7 +497,7 @@ def init():
     return parser, ds_reg, args
 
 def is_test(args):
-    return args.command == 'run' or args.command == 'run-pattern'
+    return args.command == 'run' or args.command == 'run-pattern' or args.command == 'run-file'
```
lukewrites: If this line feels like it's getting long we could use something like

```suggestion
    return args.command in ['run', 'run-pattern', 'run-file']
```

tdvt/tdvt/tdvt.py
```diff
@@ -435,27 +436,28 @@ def create_parser():
...
+    run_test_parser.add_argument('ds', help='Comma separated list of Datasource names or groups to test. See the \'list\' command.', nargs='+')
     run_test_parser.add_argument('--verify', dest='smoke_test', action='store_true', help='Verifies the connection to a data source against tests in your .ini file with SmokeTest = True.', required=False)  # noqa: E501
     run_test_parser.add_argument('--logical', '-q', dest='logical_only', help='Only run logical tests whose config file name matches the supplied string, or all if blank.', required=False, default=None, const='*', nargs='?')
     run_test_parser.add_argument('--expression', '-e', dest='expression_only', help='Only run expression tests whose config file name matches the suppled string, or all if blank.', required=False, default=None, const='*', nargs='?')
     
 
     #Run test pattern.
-    run_test_pattern_parser = subparsers.add_parser('run-pattern', help='run tests using a pattern', parents=[run_test_common_parser], usage=run_pattern_usage_text)
+    run_test_pattern_parser = subparsers.add_parser('run-pattern', help='Run single tests using a pattern.', parents=[run_test_common_parser], usage=run_pattern_usage_text)
```
lukewrites: I don't want to seem nitpicky about wording, but would changing this to "Run **a** single test using a pattern." or "Run individual tests using a pattern." add clarity?

tdvt/tdvt/tdvt.py
lriggs: fixed

tdvt/tdvt/tdvt.py
lriggs: good idea

PR #320, opened 2019-10-17
Add some documentation on running tacos

pvanderknyff:
A quick search on our docs revealed that we had no mentions of tacos beyond the pages on packaging them. Reworked some of the docs and added a page on running tacos.

Also mentioned that support links must be https.

### DIFF
--- a/_includes/docs_menu.html
+++ b/_includes/docs_menu.html
@@ -11,6 +11,7 @@
     <li><a href="{{ site.baseurl }}/docs/tdvt">Test Your Connector Using TDVT</a></li>
     <li><a href="{{ site.baseurl }}/docs/manual-test">Test Your Connector Using Manual Tests</a></li>
     <li><a href="{{ site.baseurl }}/docs/localize">Localize Your Connector</a></li>
+    <li><a href="{{ site.baseurl }}/docs/run-taco">How to Run your Packaged Connector (Taco)</a></li>
     <li><a href="{{ site.baseurl }}/docs/share">How to Run Your "Under Development" Connector</a></li>
     <li><a href="{{ site.baseurl }}/docs/package-sign">How to Package and Sign Your Connector for Distribution</a></li>
     <li>Reference</li>
--- a/_includes/docs_menu.html
+++ b/_includes/docs_menu.html
@@ -11,9 +11,9 @@
     <li><a href="{{ site.baseurl }}/docs/tdvt">Test Your Connector Using TDVT</a></li>
     <li><a href="{{ site.baseurl }}/docs/manual-test">Test Your Connector Using Manual Tests</a></li>
     <li><a href="{{ site.baseurl }}/docs/localize">Localize Your Connector</a></li>
-    <li><a href="{{ site.baseurl }}/docs/run-taco">How to Run your Packaged Connector (Taco)</a></li>
     <li><a href="{{ site.baseurl }}/docs/share">How to Run Your "Under Development" Connector</a></li>
     <li><a href="{{ site.baseurl }}/docs/package-sign">How to Package and Sign Your Connector for Distribution</a></li>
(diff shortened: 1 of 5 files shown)


_includes/docs_menu.html
```diff
@@ -11,6 +11,7 @@
     <li><a href="{{ site.baseurl }}/docs/tdvt">Test Your Connector Using TDVT</a></li>
     <li><a href="{{ site.baseurl }}/docs/manual-test">Test Your Connector Using Manual Tests</a></li>
     <li><a href="{{ site.baseurl }}/docs/localize">Localize Your Connector</a></li>
+    <li><a href="{{ site.baseurl }}/docs/run-taco">How to Run your Packaged Connector (Taco)</a></li>
```
lriggs: Putting this at the end of the list seems to fit the logical order better.

docs/index.md
```diff
@@ -24,38 +24,24 @@ See the relationship between the connector files (in blue) and the Tableau **Con
 
 ![]({{ site.baseurl }}/assets/files-overview.png)
 
-## Before you begin
+# What is a taco?
```
lukewrites: Thanks for doing this!

:27 uses `taco` and :28 uses `.taco`. We should probably standardize how and when we use each term. 

To spitball:  
`.taco` for referring to the file itself 
_taco_ to refer to a packaged connector.

What do we think about **Taco** vs **taco** for the latter case?

It may be in the docs already, but do we want to explain why it's called a taco?

And for fun I think we should include the cool .taco graphic somewhere 😉

pvanderknyff: Hmm, good question. I'll go with taco (no dot) for now. I'll also make the capilization for the headers in this file title style, in line with the other headers in other files.

d45: FYI - In our other external documentation, we use the convention of tagging the file name extensions in with `code` tags and we also include the `.` .  For example, `.trex` for Dashboard Extensions.   So you probably want to use `.taco` when you are just talking about the file. Calling Tableau Connectors "Tacos" seems pretty cool. I'd use title case for that "Taco"
Check with Kathy when she gets back.   I believe she is planning on using the taco.png file on the landing page.

pvanderknyff: Formatted the the extension names to be `.taco`, and when not talking about the file capitalized it.

docs/run-taco.md
```diff
@@ -0,0 +1,36 @@
...
+
+# Tableau Desktop
+Simply drop your `.taco` file into your `My Tableau Repository/Connectors` and launch Tableau.
+
+# Tableau Server
+1. Create a directory for Tableau connectors. For example: `C:\tableau_connectors`
+1. Drop your `.taco` file into the folder your created
+1. Set the `native_api.connect_plugins_path` option. For example:
+
+    ```
```
lukewrites: can we use `shell` or `powershell` to tag these code blocks?

docs/index.md
```diff
@@ -24,38 +24,24 @@ See the relationship between the connector files (in blue) and the Tableau **Con
...
+A `.taco` file is a packaged Tableau connector file that can be dropped into your `My Tableau Repository/Connectors` folder. They will be automatically loaded by Tableau.
+
+For more information about packaging your connector into a Taco, refer to [How to Package and Sign Your Connector for Distribution]({{ site.baseurl }}/docs/package-sign)
+
+# Before you Begin
 
 You need to do these things before you start:
 
 - Install the ODBC or JDBC driver that you’ll use with the connector you’ll create.
 - Install Tableau Desktop 2019.2 or later on a Windows or Mac computer.
```
lukewrites: Since we mention 2019.2 here, would it be useful to have a section that describes what you can do with 2019.2 vs 2019.4+, or just make explicit that the "Developer Path" option is what you can do on 2019.2/.3?

pvanderknyff:
Closing to re-open against gh-pages-dev

PR #322, opened 2019-10-17
Vendor attributes documentation 

saugatt:


docs/ui.md
```diff
@@ -27,6 +27,56 @@ These elements are defined in the manifest.xml file:
...
+        ...
+        <attr> vendor1 </attr> 
+        <attr> vendor2 </attr> 
+        <attr> vendor3 </attr> 
+
+      </attribute-list>
+      </required-attributes>
+    ...
+```
+connectionProporties.js
```
pvanderknyff: connectionProperties (typo)

docs/ui.md
```diff
@@ -27,6 +27,56 @@ These elements are defined in the manifest.xml file:
 </connector-plugin>
 ```
 
+## Vendor Attributes (Custom Fields)
+Vendors can add customized attributes to their connector plugin by using the vendor attributes.
+
+These fields have a custom label and can be used for attributes in the connection strings that are not available in the attribute list. You can currently add 3 custom fields in your connector plugin.
+
+To add a custom vendor-attribute, you will need to modify your connection-dialog.tcd, connectionResolver.tdr and connectionProporties.js file like any other attribute.
```
pvanderknyff: connectionBuilder or connectionProperties. The properties builder script is only for jdbc connectors, for odbc connectors you will be adding these to the connection builder script.

rosswbrown: Primary documentation should probably be to use them in connectionBuilder.js.  Optionally in jdbc they could also be used in a propertiesBuilder.js

docs/ui.md
```diff
@@ -27,6 +27,79 @@ These elements are defined in the manifest.xml file:
...
+Vendors can add customized attributes to their connector plugin by using the vendor attributes.
+
+These fields have a custom label and can be used for attributes in the connection strings that are not available in the attribute list. You can currently add 3 custom fields in your connector plugin.
+
+To add a custom vendor-attribute, you will need to modify your connection-dialog.tcd, connectionResolver.tdr and connectionBuilder.js.
+**For JDBC based plugins, you will need to modify connectionProperties.js instead of connectionBuilder.js**
+
+connection-resolver.tdr
+
+```
```
lukewrites: @d45 or @kpwilson are we able to format code blocks in markdown by adding the name of the language in the code block after the three backticks? e.g. ` ```xml `?

saugatt: It seems it does. Tried doing it locally. 
xmlPlatform

lukewrites: Awesome! I think it'd be worth adding those to the code blocks. Makes everything so much easier to read.

PR #326, opened 2019-10-23
Update api-reference.md

jainam1995:
Update documentation of connection builder and connection properties to use non-secure vs secure attributes

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -16,7 +16,7 @@ The Connection Resolver is made up of several components:
 
 ### connection-builder
 
-Builds the ODBC ConnectString or JDBC Connection URL.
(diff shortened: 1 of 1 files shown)


docs/api-reference.md
```diff
@@ -25,20 +25,20 @@ Builds the ODBC ConnectString or JDBC Connection URL.
...
-["DRIVER={My ODBC Driver}", "UID=myusername", "Host=myserver.somewhere.net"];
+["DRIVER={My ODBC Driver}", "Host=myserver.somewhere.net"];
 ```
 
 ---
 
 ### connection-properties
 
-Similar to connection-builder but is used to build the JDBC properties file.
+Similar to connection-builder but is used to build the JDBC properties file. For JDBC Connection URL, it is suggested that connection-properties contain secure attributes such as username and password.
```
pvanderknyff: Maybe make it more explicit that it it is only for JDBC? Also, say that connection-properties must contain X attributes, and not suggest it. This is a requirement, not a suggestion.

PR #334, opened 2019-10-24
Dev tdvt lriggs

lriggs:
Make sure failed tests re-run with the right executable.
Log process exceptions in json output.
Added ability to write unit tests for results processing.
Cleaned up some processing code.


tdvt/tdvt/config_gen/test_config.py
```diff
@@ -250,6 +250,35 @@ def build_config_name(prefix, dsname):
 def build_tds_name(prefix, dsname):
     return prefix + dsname + '.tds'
 
+class TabQueryPath(object):
+    sep = ';'
```
lukewrites: Do we want to use `os.path.sep` here to account for different seperators on different os's?

tdvt/tdvt/config_gen/test_config.py
```diff
@@ -250,6 +250,35 @@ def build_config_name(prefix, dsname):
...
+        split_paths = paths.split(TabQueryPath.sep)
+        if len(split_paths) != 3:
+            raise IndexError
+        t = TabQueryPath(split_paths[0], split_paths[1], split_paths[2])
+        return t
+
+    def to_string(self):
+        return "{0}{1}{2}{3}{4}".format(self.linux_path, self.sep, self.mac_path, self.sep, self.windows_path)
+
+    def get_tabquery_path(self, os):
```
lukewrites: Could we rename either this method or the method in `RunTimeTestConfig` to avoid confusion?

tdvt/tdvt/config_gen/test_config.py
```diff
@@ -259,25 +288,25 @@ def __init__(self, timeout_seconds=60*60, maxthread=0, d_override='', run_as_per
...
+    def set_tabquery_path_from_string(self, path_string):
+        if not path_string:
+            return
+        self.tabquery_paths = TabQueryPath.from_string(path_string)
 
     def has_customized_tabquery_path(self):
-        return self.linux_path is not '' or self.mac_path is not '' or self.windows_path is not ''
+        return self.tabquery_paths is not None
 
     def get_tabquery_path(self, os):
```
lukewrites: do we want to specify `os` to use `os.name` or `platform.system()`?

tdvt/tdvt/config_gen/test_config.py
```diff
@@ -250,6 +250,35 @@ def build_config_name(prefix, dsname):
 def build_tds_name(prefix, dsname):
     return prefix + dsname + '.tds'
 
+class TabQueryPath(object):
+    sep = ';'
+
+    def __init__(self, linux_path, mac_path, windows_path):
+        self.linux_path = linux_path
```
lukewrites: I've been wondering if it would be worth it to have TDVT start to use `Path` objects throughout. That would reduce some code complexity since `Path` has [useful methods](https://docs.python.org/3/library/pathlib.html#methods) like `.exists()` and `is_file()`. We've started doing that in the connector packager library, and it's been pretty useful I think.

tdvt/tdvt/tdvt_core.py
```diff
@@ -597,7 +612,7 @@ def run_diff(test_config, diff):
     return 0
 
 
-def run_tests_impl(test_set, test_config: TdvtInvocation):
+def run_tests_impl(test_set: TestSet, test_config: TdvtInvocation):
```
lukewrites: thanks for adding these. I have a branch with more, I'll open a PR with it.

tdvt/test/tdvt_test.py
```diff
@@ -723,6 +718,190 @@ def test_print_configuration_with_no_dsname(self):
             datasource_list.print_configurations(MockTestRegistry, None, None)
             self.assertIn(correct_out, captured_output.getvalue())
 
+class MockBatchQueueWork(tdvt_core.BatchQueueWork):
```
lukewrites: These mocks will let us do some great testing.

tdvt/tdvt/config_gen/test_config.py
lriggs: I think its a good idea. I had a bit of problem trying to store two posix paths and a windows path. I'll look into the idea of using them though.

tdvt/tdvt/config_gen/test_config.py
lriggs: Updated these.

PR #341, opened 2019-10-28
add changelog to packaging

lukewrites:


README.md
```diff
@@ -22,15 +28,15 @@ Review the [Tableau Connector SDK developer guide](https://tableau.github.io/con
 
 # Samples
 
-There are two [standalone Postgres example connectors](https://github.com/tableau/connector-plugin-sdk/tree/master/samples/plugins) that use ODBC and JDBC.
+We have created three [standalone example connectors](https://github.com/tableau/connector-plugin-sdk/tree/master/samples/plugins) that use ODBC and JDBC.
```
pvanderknyff: Minor, minor nitpick - maybe say "... several standalone example connectors..." or something similar so we don't have to update this if we add more samples in the future.

lukewrites: Good suggestion, thanks 👍

PR #345, opened 2019-10-29
Fix issue with logging

pvanderknyff:
One of the imported modules must have set the root logger and that messed our logging up. Got it to a working state, tested against 3.7.4. Had to move the initialization of the logger to package.py from helper.py as part of the fix.

Opening as a hotfix against master, since I'm not sure if we've cleaned up dev. Will also merge into dev once it's merged into master.

### DIFF
--- a/connector-packager/connector_packager/helper.py
+++ b/connector-packager/connector_packager/helper.py
@@ -1,14 +1,13 @@
 import os
 import logging
 
-from imp import reload
 from pathlib import Path
 
 from .version import __version__
 
 
 PATH_ENVIRON = "PATH"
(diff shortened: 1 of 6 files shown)


jsun28:
still didn't understand the cause of this problem, what does "One of the imported modules must have set the root logger" mean? the root logger is set by calling init_logging function, right? why does the location of init_logging matter?

lukewrites:
Lgtm could you also bump the version of connector-packager?

pvanderknyff:
@lukewrites I'll update the version number in a separate PR, since this is tecnically a hotfix.

PR #354, opened 2019-11-04
describe use of .password file

lukewrites:
Change our docs to instruct users to use a `.password` file rather than hard coding the pw in their `.tds` files.


lukewrites:
Should we retain the documentation about hard coding it?

lukewrites:
Oh I just saw yours! I can drop the manual steps into your branch, but yours looks great.

PR #357, opened 2019-11-04
Revert merges that should be in vNext

lukewrites:
Reverting some merges into dev that should be in vNext.


lukewrites:
@lriggs reverting the merge into dev won't affect what's in our git history. My plan is to open a branch off of clean `dev` and then cherry-pick in the commits. If worst comes to worst I'll just recreate them. They're still around [in our Pull requests tab](https://github.com/tableau/connector-plugin-sdk/pulls?q=is%3Apr+is%3Aclosed) after the revert.

PR #360, opened 2019-11-05
Add deprecated api doc

pvanderknyff:
Putting this in the 2020.1 holding branch for docs.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -221,7 +221,7 @@ Example:
 ## Deprecated API
 
 ### SetImpersonateAttributes connection helper
-This connection helper is deprecated as of Tableau 2020.1, since we always set impersonate attributes for all connectors. Trying to use this in a javascript component will throw an error when attempting to connect.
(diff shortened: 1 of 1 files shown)


docs/api-reference.md
```diff
@@ -218,3 +218,10 @@ Example:
 
     formattedParams.push(connectionHelper.FormatKeyValuePair(driverLocator.keywordDriver, driverLocator.LocateDriver(attr)));
 
+## Deprecated API
+
+### SetImpersonateAttributes connection helper
+This connection helper is deprecated as of Tableau 2020.1, since we always set impersonate attributes for all connectors. Trying to use this in a javascript component will throw an error when attempting to connect.
```
lukewrites:
```suggestion
This connection helper is deprecated as of Tableau 2020.1, since we always set impersonate attributes for all connectors. Trying to use this in a JavaScript component will throw an error when attempting to connect.
```

lukewrites: sorry for the nit, but I _think_ we're standardizing around capitalizing JavaScript like that.

PR #364, opened 2019-11-07
Update run_file to check if file exists

lukewrites:
Makes the following changes:
  * A Path object is created with args.run_file
  * TDVT checks if the file exists before continuing.
  * output_dir is also a Path object
  * the run_file() fn is updated to work with Path objects.
  * enqueue_failed_tests() is updated to work with Path objects.


lukewrites:
It should be in dev, forgot and branched off master. 😢

PR #365, opened 2019-11-07
Update run_file to check if file exists

lukewrites:
Makes the following changes:
  * A Path object is created with `args.run_file`
  * TDVT checks if the file exists before continuing.
  * `output_dir` is also a Path object
  * the `run_file()` fn is updated to work with Path objects.
  * `enqueue_failed_tests()` is updated to work with Path objects.

### DIFF
--- a/tdvt/tdvt/tdvt.py
+++ b/tdvt/tdvt/tdvt.py
@@ -20,17 +20,18 @@ import shutil
 import subprocess
 import threading
 import time
-from .version import __version__
+from pathlib import Path
 from typing import List
 
-from .config_gen.gentests import list_configs, list_config
 from .config_gen.datasource_list import print_ds, print_configurations, print_logical_configurations
+from .config_gen.gentests import list_configs, list_config
+from .config_gen.tdvtconfig import TdvtInvocation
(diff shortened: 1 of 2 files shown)


lukewrites:
@pvanderknyff based off `dev` _and_ I updated the changelog 😉

tdvt/tdvt/tdvt.py
```diff
@@ -697,12 +698,17 @@ def main():
...
-            output_dir = os.getcwd()
+            rfile = Path(args.run_file)
+            output_dir = Path.cwd()
             max_threads = get_level_of_parallelization(args)
-            sys.exit(run_file(args.run_file, output_dir, max_threads, args))
+            if rfile.is_file():
+                sys.exit(run_file(rfile, output_dir, max_threads, args))
+            else:
+                logging.error("Unable to open " + rfile.__str__())
+                return -1
```
lriggs: I don't know if it matters but the other branches are doing sys.exit

lukewrites: good point, thanks, I'll keep it consistent.

PR #371, opened 2019-11-12
Create pull_request_template.md

lukewrites:
I thought it might be useful to always show some guidelines when a PR is opened. The text would show up at the top of any Pull Request description when you open it and can be removed.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,7 @@
 **A Helpful Guide For Navigating Our Branches**
 
 * If this is a hotfix, make your PR against `master`.
-* If your change can ship immediately, make your PR against `dev`.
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,11 @@
+**A Helpful Guide For Navigating Our Branches**
+
+* If this is a hotfix, make your PR against `master`.
+* If your change can ship immediately, make your PR against `dev`.
```
lriggs: or `dev-tdvt`

lukewrites: Changed it. I wonder if we'd like to have `dev-sdk` and `dev-packaging` as well? We could do away with the general `dev` branch in that case, and it'd be clear where things should go/where to branch off of.

PR #376, opened 2019-11-19
Move CI to GitHub Actions

lukewrites:
Run TDVT & Connector Packager unit tests via GitHub Actions.

This runs a subset of the TDVT unit tests (the ones that don't require tabquerycli) and all packaging tests.

The tests are run across Linux/Mac/Windows on Python 3.7 and JDK 1.8/1.10/1.12

image


lukewrites:
> 
> 
> Should this be in dev-tdvt or master? Not sure how the yml files work with our branch model.

We'd want it to go against master then bring it in to the different branches since we want the CI to be able to run everywhere. I'm playing with this to try to get the packager tests to run as well.

lukewrites:
CI shows four unit tests failing, which is the same as I have on my local.

sonia0504:
opened bug#1033597 to track these failures

PR #379, opened 2019-11-20
Fix MySQL Sample ssl

pvanderknyff:
Our sample connector did not disable ssl properly. Since it was trivial, added the ssl checkbox. Removed the 5.3 driver from the driver resolver since it'd need different logic for the ssl mode parameters and it is past EOL.

### DIFF
--- a/samples/plugins/mysql_odbc/connectionResolver.tdr
+++ b/samples/plugins/mysql_odbc/connectionResolver.tdr
@@ -13,6 +13,7 @@
           <attr>username</attr>
           <attr>password</attr>
           <attr>port</attr>
+          <attr>sslmode</attr>
           <attr>odbc-connect-string-extras</attr>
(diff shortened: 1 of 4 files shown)


samples/plugins/mysql_odbc/connectionResolver.tdr
```diff
@@ -13,6 +13,7 @@
           <attr>username</attr>
           <attr>password</attr>
           <attr>port</attr>
+          <attr>sslmode</attr>
```
sxlin: Have you tested connection with SSL enabled?  I think sslcert should be required too.

pvanderknyff: I have, it works and I confirmed with wireshark that the traffic is encrypted. No cert requried.

PR #383, opened 2019-11-22
Change connectionconfig's children could be in random order

jsun28:
Change connectionconfig's children could be in random order.


lukewrites:
LGTM, thanks for updating the Changelog!

PR #388, opened 2019-11-25
Fix api doc in 2019.4

liyunbao:
**A Helpful Guide For Navigating Our Branches**

* If this is a hotfix, make your PR against `master`.
* If your change can ship immediately, make your PR against `dev-tdvt`.
* If your change is for a future release, make your PR against the appropriate holding branch (eg `dev-2020.1`)

_Also:_
* Please update the changelog of the project you modified, either under `vNext` or under a point release if you bumped the version.
* If you are merging into master, update the root `README.md` if the PR includes a release of the SDK, Connector Packager, or TDVT.

Thank you and feel free to delete this from your PR description!

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -51,10 +51,18 @@ Similar to connection-builder but is used to build the JDBC properties file. For
 {"server" : "myserver.somewhere.net", "username" : "myusername", "password" : "mypassword"}
 ```
 
-**Return:** a dictionary of key/value pairs that will be written to the properties file
+**Return:** dictionary or vector
+
+Recommended: a dictionary of key/value pairs that will be written to the properties file
 
 ```javascript
 ["UID" : "myusername", "Host" : "myserver.somewhere.net", "PWD" : "mypassword"];
+```
+
+Since 2019.4.1: vector of formatted key=value pairs that will be written to the properties file
+
+```javascript
+["UID=myusername", "Host=myserver.somewhere.net", "PWD=mypassword"];
 ```
 
 ---
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -53,13 +53,13 @@ Similar to connection-builder but is used to build the JDBC properties file. For
 
 **Return:** dictionary or vector
 
-Recommended: a dictionary of key/value pairs that will be written to the properties file
+a dictionary of key/value pairs that will be written to the properties file(recommended and available since 2019.4.1)
 
 ```javascript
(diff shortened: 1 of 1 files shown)


lriggs:
What about documenting both APIs but say 'Recommended' for the new one and 'Since 2019.4.1' or something like that.

pvanderknyff:
> What about documenting both APIs but say 'Recommended' for the new one and 'Since 2019.4.1' or something like that.

Sounds good to me. That gives authors enough information to make the right choice.

liyunbao:
I updated the doc based on the discussion. Let me know if any changes are needed. @pvanderknyff @lriggs

docs/api-reference.md
```diff
@@ -51,7 +51,15 @@ Similar to connection-builder but is used to build the JDBC properties file. For
...
-**Return:** vector of formatted key=value pairs that will be written to the properties file
+**Return:** dictionary or vector
+
+Recommended: a dictionary of key/value pairs that will be written to the properties file
+
+```javascript
+["UID" : "myusername", "Host" : "myserver.somewhere.net", "PWD" : "mypassword"];
+```
+
+Since 2019.4.1: vector of formatted key=value pairs that will be written to the properties file
```
lriggs: Isn't the dictionary the new return type for 2019.4.1+ ?

liyunbao: Right. I think I made a mistake here. How about I change the description as below? Do you think that makes sense?

a dictionary of key/value.... (recommended and available since 2019.4.1)
...
vector of formatted key=value pairs ...
...

lriggs: Looks great.

liyunbao:
new patch uploaded

docs/api-reference.md
```diff
@@ -51,7 +51,15 @@ Similar to connection-builder but is used to build the JDBC properties file. For
...
-**Return:** vector of formatted key=value pairs that will be written to the properties file
+**Return:** dictionary or vector
+
+a dictionary of key/value pairs that will be written to the properties file(recommended and available since 2019.4.1)
+
+```javascript
+["UID" : "myusername", "Host" : "myserver.somewhere.net", "PWD" : "mypassword"];
+```
+
+vector of formatted key=value pairs that will be written to the properties file
```
pvanderknyff: Adding "Option 1:" at the beginning of line 56 and "Option 2" at the beginning of line 62 would make this more clear

pvanderknyff: Mention that this option is not recommended and only included for backwards compatibility.

PR #393, opened 2019-12-04
Keep all .txt and .log files including jdbc server log.

lriggs:
Keep all .txt and .log files including jdbc server log. Remove unneeded imports. Actually compress the log zip file.

Standardize help descriptions.

### DIFF
--- a/tdvt/tdvt/tdvt.py
+++ b/tdvt/tdvt/tdvt.py
@@ -372,49 +372,44 @@ list_usage_text = '''
     
 run_usage_text = '''
     The 'run' argument can take a single datasource, a list of data sources, or a test suite name in any combination.
-        run vertica
-        run sqlserver,vertica
-        run standard
+        run postgres_odbc,postgres_jdbc
 
     The 'run' argument can also take the --verify flag to run a connection test against tests with SmokeTest = True set.
-        run postgres --verify
+        run postgres_odbc --verify
 
     Both logical and expression tests are run by default.
-    Run all sqlserver expression tests
-       run -e sqlserver
+    Run all expression tests
+       run postgres_odbc -e
 
-    Run all vertica logical tests
-        run -q vertica
+    Run all logical tests
+        run postgres_odbc -q
 
     There are multiple suites of expression tests, for example, standard and LOD (level of detail). The config files that drive the tests
     are named expression_test.sqlserver.cfg and expression.lod.sqlserver.cfg.
     To run just one of those try entering part of the config name as an argument:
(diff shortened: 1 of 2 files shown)


tdvt/tdvt/tdvt.py
```diff
@@ -373,49 +368,44 @@ def get_level_of_parallelization(args):
...
 run_usage_text = '''
     The 'run' argument can take a single datasource, a list of data sources, or a test suite name in any combination.
-        run vertica
-        run sqlserver,vertica
-        run standard
+        run postgres_odbc,postgres_jdbc
 
     The 'run' argument can also take the --verify flag to run a connection test against tests with SmokeTest = True set.
-        run postgres --verify
+        run postgres_odbc --verify
```
lukewrites: these are good changes, thanks

tdvt/tdvt/tdvt.py
```diff
@@ -102,12 +97,12 @@ def copy_files_to_zip(self, dst_file_name, src_dir, is_logs):
...
             log_dir = os.path.join(src_dir, optional_dir_name)
-            glob_path = glob.glob(os.path.join(log_dir, 'log*.txt'))
-            glob_path.extend(glob.glob(os.path.join(log_dir, 'tabprotosrv*.txt')))
+            glob_path = glob.glob(os.path.join(log_dir, '*.txt'))
+            glob_path.extend(glob.glob(os.path.join(log_dir, '*.log')))
             glob_path.extend(glob.glob(os.path.join(log_dir, 'crashdumps/*')))
         else:
             glob_path = glob.glob(os.path.join(src_dir, 'actual.*'))
-        with ZipFile(dst, mode) as myzip:
+        with zipfile.ZipFile(dst, mode, zipfile.ZIP_DEFLATED) as myzip:
```
lukewrites: From [the Python docs](https://docs.python.org/3/library/zlib.html#module-zlib), I see that `zipfile.ZIP_DEFLATED` uses the `zlib` module, which in turn has an external dependency on `zlib`. Don't think that'll be an issue, but just something to have in the back of our heads if anyone complains about it.

lukewrites:
Oh, it'd be great if we could add these changes to the **Unreleased** section of `tdvt/CHANGELOG.md`

tdvt/tdvt/tdvt.py
lriggs: I was a bit concerned about that too. I tested on windows, mac and linux and we at least have it installed there. If we have problems we could do a try/catch with the import statement and either compress or not based on that.

PR #405, opened 2019-12-13
Dev tdvt lriggs

lriggs:
Check for tabquery path before running unit tests that need it.
Cache list of tests for test_set. Fixed bug where run-pattern would take several seconds to run 0 tests.

### DIFF
--- a/tdvt/tdvt/config_gen/test_config.py
+++ b/tdvt/tdvt/config_gen/test_config.py
@@ -65,7 +65,7 @@ class TestSet(object):
            Return the sorted list of tests.
 
         """
-        if self.test_list_checked == True:
+        if self.test_list_checked:
             return self.test_list_cached
 
         final_test_list = self.__generate_test_file_list()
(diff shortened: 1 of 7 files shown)


tdvt/tdvt/config_gen/test_config.py
```diff
@@ -62,24 +65,20 @@ def generate_test_file_list(self):
...
+        self.test_list_cached = final_test_list
+        self.test_list_checked = True
+        return self.test_list_cached
 
     def get_test_dirs(self):
         return (self.root_dir, get_local_test_dir())
 
-    def generate_test_file_list_from_config(self):
-        """Read the config file and generate a list of tests."""
+    def __generate_test_file_list(self):
```
lukewrites: Since we've got a few methods called `generate_test_file_list` now, it'd be great if we could add type hints to them.

tdvt/tdvt/tabquery_path.py
```diff
@@ -0,0 +1,23 @@
...
+    def from_array(paths):
+        if len(paths) != 3:
+            raise IndexError
+        t = TabQueryPath(paths[0], paths[1], paths[2])
+        return t
+
+    def to_array(self):
+        return [self.linux_path, self.mac_path, self.windows_path]
+
+    def get_path(self, os):
```
lukewrites: Do we gain anything by requiring an `os` arg? The function could just be something like
```python
def get_path(self):
   if sys.platform.startswith('darwin'):
        ...
```

lriggs: Not really. It's used for testing. I'll think about changing it.

PR #417, opened 2019-12-21
Postgres jdbc kerberos plugin

devanshsoni9:
This PR is right now intended for comments/feedback on the sample plugin. I won't merge until before the deadline for 2020.2 or advised otherwise. I discussed with Ross and his idea was to include this plugin in 2020.2 only if the Postgres JDBC 42.2.10 driver is released before that. This is because SSPI won't work until the [fix](https://github.com/pgjdbc/pgjdbc/pull/1651) for issue [#1482](https://github.com/pgjdbc/pgjdbc/issues/1482) is released in the new driver jar.

### DIFF
--- a/samples/plugins/postgres_jdbc_kerberos/connection-dialog.tcd
+++ b/samples/plugins/postgres_jdbc_kerberos/connection-dialog.tcd
@@ -1,0 +1,12 @@
+<connection-dialog class='postgres_jdbc_kerberos'>
+  <connection-config>
+    <authentication-mode value='ComboBoxIntegrated'/>
+    <authentication-options>
+      <option name="UseWindows" value="auth-integrated" />
+      <option name="UsernameAndPassword" value="username-password" />
+    </authentication-options>
+	<db-name-prompt value="Database: " />
+    <port-prompt value="Port: " default="5432" />
+    <show-ssl-checkbox value="true" />
+  </connection-config>
+</connection-dialog>
--- a/samples/plugins/postgres_jdbc_kerberos/manifest.xml
+++ b/samples/plugins/postgres_jdbc_kerberos/manifest.xml
@@ -1,0 +1,38 @@
+<?xml version='1.0' encoding='utf-8' ?>
+
+<connector-plugin class='postgres_jdbc_kerberos' superclass='jdbc' plugin-version='0.0.0' name='Postgres JDBC with Kerberos' version='18.1' min-version-tableau='2019.4'>
+  <vendor-information>
+      <company name="Company Name"/>
+      <support-link url = "http://example.com"/>
+  </vendor-information>
+  <connection-customization class="postgres_jdbc_kerberos" enabled="true" version="10.0">
+    <vendor name="vendor"/>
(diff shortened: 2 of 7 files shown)


lukewrites:
@devanshsoni9 I changed the base of this PR to the 2020-2 branch. if you're OK with it going in there you can mark it Ready for review.

samples/plugins/postgres_jdbc_kerberos/manifest.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version='1.0' encoding='utf-8' ?>
+
+<connector-plugin class='postgres_jdbc_kerberos' superclass='jdbc' plugin-version='0.0.0' name='Postgres JDBC with Kerberos' version='18.1' min-version-tableau='2019.4'>
```
jainam1995: Do you want to change the min version to 2020.2 if its only going to come in after that release ?

samples/plugins/postgres_jdbc_kerberos/connection-dialog.tcd
```diff
@@ -0,0 +1,12 @@
+<connection-dialog class='postgres_jdbc_kerberos'>
+  <connection-config>
+    <authentication-mode value='ComboBoxIntegrated'/>
+    <authentication-options>
+      <option name="UseWindows" value="auth-integrated" />
+      <option name="UsernameAndPassword" value="username-password" />
+    </authentication-options>
+	<db-name-prompt value="Database: " />
```
jainam1995: indendation

samples/plugins/postgres_jdbc_kerberos/connection-dialog.tcd
```diff
@@ -0,0 +1,12 @@
+<connection-dialog class='postgres_jdbc_kerberos'>
+  <connection-config>
+    <authentication-mode value='ComboBoxIntegrated'/>
+    <authentication-options>
+      <option name="UseWindows" value="auth-integrated" />
```
devanshsoni9, line 5: @rosswbrown should we be using `auth-kerberos` in instead of `auth-integrated` ? 
Also, should we be documenting kerberos in the [supported auth modes](https://tableau.github.io/connector-plugin-sdk/docs/auth-modes)

devanshsoni9: @sidwray same question to you about documentation

rosswbrown: From what I can tell in the code `auth-kerberos` happens to be some teradata specific implementation detail and doesn't have broad platform meaning.   I think this is correct, but I defer to you as person authoring/testing this.  We should probably consider changing this sample over to using modular connection dialog definition, since this is a forward looking sample.  This notion of which auth to use simplifies greatly with that style.

I don't believe we document anything until we're planning to ship this feature, which I thought was post 2020.2.

devanshsoni9:
will not ship in 2020.2 in favor of using modular connection dialog

PR #418, opened 2019-12-23
TFSID 1040094: Add in new tdvt test files to test snowflake datename does not truncate

mirhyman:
I changed the MON dialect to MMMM so that we return "Decemeber" instead of "Dec" when using TO_CHAR(datename)

see: Full name is the convention we document for Tableau calcs: https://help.tableau.com/current/pro/desktop/en-us/functions_functions_typeconversion.htm.  MMMM is the correct one to use here over MON and https://docs.snowflake.net/manuals/sql-reference/functions-conversion.html#label-date-time-format-conversion

Added in tests to snowflake to test that we are not truncating months in the datename function


mirhyman:
@lukewrites @pvanderknyff  requesting review if possible

sxlin:
Looks good to me.

PR #437, opened 2020-01-10
Add string.contains.regex test

pvanderknyff:
Tests the escaping of regex special characters in the contains function, this would have caught the databricks bug.

### DIFF
--- a/tdvt/CHANGELOG.md
+++ b/tdvt/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Add string.contains.regex test to test escaping regex special characters in the Contains function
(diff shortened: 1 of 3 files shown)


pvanderknyff:
Spark, postgres, Impala

lukewrites:
Oh, I should have asked before, do you want to do a release with this new testcase? If that's the plan, go ahead and bump the version, changelog, and readme with this PR, `dev-tdvt` is clean.

pvanderknyff:
Yes, I do, since I'll need to do a bit or work to exclude this test on any datasource it fails on (like Impala). I'll update everything.

pvanderknyff:
Not sure since I can't run this test on StandardDBs until I make a release. It passes my random selection of them with the exception of Impala, which is hive-based but has it's own dialect.

lukewrites:
if it's going to take some finetuning on local you can bump the version to something like 2.x.xb1 and then just package it and upload to artifactory

pvanderknyff:
Confirmed test passes for all in standard suite, merging

PR #438, opened 2020-01-10
Packaging jshah

pvanderknyff:
This didn't get merged into dev, but it looks like the branch was never deleted, so I can just merge that into dev.


jainam1995:
A small query : Wouldnt the correct approach be to merge packaging branch (https://github.com/tableau/connector-plugin-sdk/tree/packaging) which has this change to master ?

pvanderknyff:
> A small query : Wouldnt the correct approach be to merge packaging branch (https://github.com/tableau/connector-plugin-sdk/tree/packaging) which has this change to master ?

I think there might be conflicts. The packaging branch is basically deprecated at this point, and there may be commits we don't really want, or commits that are already there. Probably why this change never got picked up.

rosswbrown:
Can we remove the packaging branch after this change is merged?  Maybe some other branch cleanup if possible.

lukewrites:
> 
> 
> Can we remove the packaging branch after this change is merged? Maybe some other branch cleanup if possible.

good idea.

PR #445, opened 2020-01-11
Pvanderknyff/sample unification

pvanderknyff:
Proposal to unify our samples so that they provide a consistent experience. 

- Display names now have "Sample" in front of them. This way, it's clear they are not a "real" connector and they appear next to each other on the connector list.
- Support link now points to https://tableau.github.io/connector-plugin-sdk/. If, somehow, someone actually nees support with one of these, that's as good a palce as any.
- Added the actual place to download the driver as the driver download link, except for SQLite, as the site to download that does not have https.


lukewrites:
I think @rosswbrown has a good point, it might be nice to just leave them pointing to example.com. 

_Or_ we could point them to a custom page that says "Your connector plugin's support link works! Now, you need to change the URL in `manifest.xml` to one that points to your own documentation."

PR #449, opened 2020-01-24
Merge gh-pages-dev into gh-pages

pvanderknyff:
Some things have been gathering dust in the dev branch for the documentation.


docs/api-reference.md
```diff
@@ -51,7 +51,15 @@ Similar to connection-builder but is used to build the JDBC properties file. For
 {"server" : "myserver.somewhere.net", "username" : "myusername", "password" : "mypassword"}
 ```
 
-**Return:** vector of formatted key=value pairs that will be written to the properties file
+**Return:** dictionary or vector
```
lukewrites: just for the sake of clarity, do we want to also include a sample js object that might be returned?

docs/api-reference.md
```diff
@@ -51,7 +51,15 @@ Similar to connection-builder but is used to build the JDBC properties file. For
...
+
+a dictionary of key/value pairs that will be written to the properties file(recommended and available since 2019.4.1)
+
+```javascript
+["UID" : "myusername", "Host" : "myserver.somewhere.net", "PWD" : "mypassword"];
+```
+
+vector of formatted key=value pairs that will be written to the properties file
```
lukewrites: More of a general question about this doc page. We use some non-JS terms here (vector for array, dictionary for object) and I have always wondered why. Would it cause any harm to remove them?

lriggs: seems like a good idea to use the JS terminology.

docs/api-reference.md
pvanderknyff: The postgres jdbc sample has been updated to return the dictionary

PR #456, opened 2020-02-04
Add support for new dialog files in packager

pvanderknyff:
Adds XSD files for modular connection dialogs as well. Will add some unit tests as well.

Opening this against dev-2020.2 since this won't be able to be used until that release.

### DIFF
--- a/connector-packager/connector_packager/version.py
+++ b/connector-packager/connector_packager/version.py
@@ -1,2 +1,2 @@
 __version__ = '1.0.0'
-__min_version_tableau__ = '2019.4'
+__default_min_version_tableau__ = '2019.4'
--- a/connector-packager/connector_packager/xml_parser.py
+++ b/connector-packager/connector_packager/xml_parser.py
@@ -153,6 +153,19 @@ class XMLParser:
                 # add new file to list
                 self.file_list.append(new_file)
 
+                # If connection-metadata, make sure that connection-fields file exists
+                if child.tag == "connection-metadata":
+                    connection_metadata_exists = False
+
+                    for xml_file in self.file_list:
+                        if xml_file.file_type == "connection-fields":
+                            connection_metadata_exists = True
+                            break
+
+                    if not connection_metadata_exists:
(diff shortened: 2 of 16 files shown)


connector-packager/connector_packager/xml_parser.py
```diff
@@ -153,6 +153,19 @@ def parse_file(self, file_to_parse: ConnectorFile) -> bool:
                 # add new file to list
                 self.file_list.append(new_file)
 
+                # If connection-metadata, make sure that connection-fields file exists
+                if child.tag == "connection-metadata":
```
rosswbrown: We're inconsistent about single vs double quote, seems like mostly single, match the file I suppose.

connector-packager/connector_packager/xml_parser.py
```diff
@@ -153,6 +153,19 @@ def parse_file(self, file_to_parse: ConnectorFile) -> bool:
...
+                    connection_metadata_exists = False
+
+                    for xml_file in self.file_list:
+                        if xml_file.file_type == "connection-fields":
+                            connection_metadata_exists = True
```
rosswbrown: connection_fields_exists

connector-packager/connector_packager/xml_parser.py
pvanderknyff: Changed to single quote

connector-packager/connector_packager/xml_parser.py
pvanderknyff: Changed

connector-packager/connector_packager/version.py
```diff
@@ -1,2 +1,2 @@
 __version__ = '1.0.0'
-__min_version_tableau__ = '2019.4'
+__default_min_version_tableau__ = '2019.4'
```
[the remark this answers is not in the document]
pvanderknyff: With my changes, this is the default. Most connectors at the moment can work with 2019.4. I added a method that determines the minimum version, and if there are files that use the new modular connection dialog framework we stamp 2020.2 instead.

PR #468, opened 2020-03-06
Remove option to sign taco using packager

pvanderknyff:
Instead of having our packager to call jarsigner from python, we want users to call jarsigner directly. This means we do not have to match the interface of jarsigner in the packager, and means partners can use external docuemntation for signing jar files to sign the taco file. This pull request must be matched with a documentation change directing partners to the official jarsigner documentation.


connector-packager/connector_packager/package.py
```diff
@@ -125,20 +115,7 @@ def main():
...
-    alias_from_args = args.alias
-    keystore_from_args = args.keystore
-
-    if not validate_signing_input(package_dest_path, package_name, alias_from_args, keystore_from_args):
-        logger.debug("Signing input validation failed. check " + str(log_file) + " for more information.")
-        return
-
-    if not jdk_sign_jar(package_dest_path, package_name, alias_from_args, keystore_from_args):
-        logger.info("Signing failed. check console output and " + str(log_file) + " for more information.")
-        return
```
rosswbrown: Can signing files also be removed?

lukewrites:
Since this is a breaking change we need to increment the version.

PR #471, opened 2020-03-11
Pvanderknyff/2020 1

pvanderknyff:
A bunch of changes that were hanging out in the 2020.1 branch that should be merged now that that version has been released. Looks to be updating some samples to use vendor attributes and adding odbc connect string extras.


saugatt:
LGTM

PR #472, opened 2020-03-11
Pvanderknyff/docs 2020 1

pvanderknyff:
The 2020.1 changes to the documentation. Mainly a couple of stray typos being fixed and Saugat's vendor attributes documentation.

### DIFF
--- a/docs/ui.md
+++ b/docs/ui.md
@@ -42,9 +42,9 @@ connection-resolver.tdr
       <required-attributes>
       <attribute-list>
         ...
-        <attr> vendor1 </attr> 
-        <attr> vendor2 </attr> 
-        <attr> vendor3 </attr> 
+        <attr> vendor1 </attr>
+        <attr> vendor2 </attr>
+        <attr> vendor3 </attr>
 
       </attribute-list>
       </required-attributes>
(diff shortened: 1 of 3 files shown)


pvanderknyff:
> Just in case, this is not in your changes, but thought you might want to take a quick look at the 'Connection properties' section of api-reference.md. Looks as below, suggesting two return values or a missing word or two?

This is correct. There are two possible return values. One of them is a vector of values where "key=value" that we then parsed in c++, which broke if any of the values had an equal sign in it. We added the dictionary (or, in javascript terminology, object) as a return type to mitigate that.

docs/ui.md
```diff
@@ -27,10 +27,83 @@ These elements are defined in the manifest.xml file:
...
+            ...
+            <vendor1-prompt value="Log Level: "/>
+            <vendor2-prompt value="Protocol Version: "/>
+            <vendor3-prompt value="Char Set: "/>
+
+        </connection-config>
+      </connection-dialog>
+```
+
+connectionBuilder.js (Non-JDBC)
```
rosswbrown: Non-JDBC is a little confusing, its ODBC.  We can do this later, but really we have ODBC and JDBC based connectors only right now.  We might consider structuring the docs around ODBC, JDBC, future, etc eventually.

pvanderknyff: Changed this to ODBC

PR #474, opened 2020-03-13
Add version disclaimer

pvanderknyff:
Make it clear that the SDK is currently targeting 2020.1. In the future, we should look at making this process automatic and transparent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 This project consists of documentation, example files, the Tableau Datasource Verification Tool (TDVT) test harness, and a packaging tool that you can use to build and customize a Tableau Connector that uses an ODBC or JDBC driver.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,8 +4,11 @@
 
 This project consists of documentation, example files, the Tableau Datasource Verification Tool (TDVT) test harness, and a packaging tool that you can use to build and customize a Tableau Connector that uses an ODBC or JDBC driver.
 
+The latest version of the SDK is always targeted towards the latest, non-beta version of Tableau. Right now, this is **Tableau Desktop/Server 2020.1**. This means that the samples may not work on older versions of tableau, and connectors packaged with newer versions may not work in older versions.
```
joe-langford: tableau -> Tableau
consider "with newer versions may not work in older versions" -> "with newer versions of the SDK may not work in older versions of Tableau"

pvanderknyff: Updated, thanks!

PR #481, opened 2020-03-23
Field name attribute documentation 

rosswbrown:
Modular connection dialog field name attribute documentation.

This is the minimal set of name and value field information for platform integration:
- Endpoint
- SSL
- Authentication

This is the most concise way I found to communicate the core platform requirements.  I did not try to document the entire set of reserved connection attributes and values that was compiled here: https://gitlab.tableausoftware.com/connectivity/connector-plugin-sdk-internal/blob/master/docs/attributes.md

I'd be open to people's thoughts on the best way to communicate the entire set of reserved attributes.  I'll open anything like that in a follow-up change.

### DIFF
--- a/docs/mcd.md
+++ b/docs/mcd.md
@@ -46,7 +46,7 @@ Each connection attribute is represented by a field element in the XML. The fiel
 
 | Name  | Meaning | Optional? | Value Notes | Other Notes |
 | ----  | ------- | --------- | ----------- | ----------- |
-| name  | Unique name of the field, which is used in the ConnectionBuilder() | No | Names must be unique, and cannot be any of the reserved names `dbname`, `schema`, `tablename`. | If there is a Tableau-defined name for this attribute, that name must be used. The Tableau-defined attribute name list is coming soon. |
+| name  | Unique name of the field: used in the platform, connection-normalizer and connection-builder | No | Names must be unique, and cannot be any of the reserved names `dbname`, `schema`, `tablename`. | If there is a Tableau-defined name for this attribute, that name must be used. See [Connection Field Platform Integration](##Connection-Field-Platform-Integration) section below. |
 | label | Label that appears on the connection dialog for the field | No | | |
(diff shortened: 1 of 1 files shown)


docs/mcd.md
```diff
@@ -182,8 +225,6 @@ The images show the Connection Dialog produced using the Connection Fields file
     </conditions>
   </field>
  
-  <field name="AutoReconnect" label="" value-type="string" default-value="0" editable="false" />
```
rosswbrown: This was intentional removal

sxlin: It would be great if you can add sslmode in the example.

rosswbrown: I'll add a note to do as follow-up in the TFS item.  I'll need to update the screenshots if I change examples here.

rosswbrown: Instead of modifying existing example I opted to make a clean boolean focused Example 3.

docs/mcd.md
```diff
@@ -114,8 +114,51 @@ The connector will not load if there are circular references.
 | field | Name of the field whose value will be checked for equality | No | Any string value | |
 | value | The value to match. If the field's value is equal to this, the condition is true. | No | Any string value | |
 
+## Connection Field Platform Integration 
 
-## Example - A Non-Editable Field
+As referenced above in the `<field>` element section, some `name` attribute values describe platform functionality. It is important that if the connector functionality desired matches any of the descriptions below then names and values below must be used.
```
joe-langford: nits: consider 
"connector functionality desired" -> "desired connector functionality"
"then names" -> "then the names"

docs/mcd.md
```diff
@@ -114,8 +114,51 @@ The connector will not load if there are circular references.
...
+
+Additionally there are a set of reserved `name` attribute values not documented at this time. Recommendations, documentation and enforcement coming soon.
+
+### Endpoint
+
+The endpoint attributes describe the unique parameters of a connection. Many connections provider additional field names and values not defined by the platform.
```
joe-langford: provider -> provide

PR #492, opened 2020-04-07
Enforce platform reserved Connection Field name values 

rosswbrown:
See documentation in 'Connection Field Platform Integration' section of [Connection Dialog V2](https://tableau.github.io/connector-plugin-sdk/docs/mcd) 

Enforce allowed names: server, port, sslmode, authentication, username, password or prefix v-

More details in TFS1092438.

### DIFF
--- a/connector-packager/tests/test_xsd_validator.py
+++ b/connector-packager/tests/test_xsd_validator.py
@@ -65,3 +65,20 @@ class TestXSDValidator(unittest.TestCase):
         logging.debug("test_validate_single_file xml violations:")
         for violation in xml_violations_buffer:
             logging.debug(violation)
+
+    def test_validate_vendor_prefix(self):
+
+        test_file = TEST_FOLDER / Path("modular_dialog_connector/connectionFields.xml")
+        file_to_test = ConnectorFile("connectionFields.xml", "connection-fields")
+        xml_violations_buffer = []
+
+        self.assertTrue(validate_single_file(file_to_test, test_file, xml_violations_buffer),
+                        "Valid XML file not marked as valid")
+
+        test_file = TEST_FOLDER / Path("broken_xml/connectionFields.xml")
+        self.assertFalse(validate_single_file(file_to_test, test_file, xml_violations_buffer),
+                         "XML file that doesn't follow valid name values marked as valid")
+
+        logging.debug("test_validate_vendor_prefix xml violations:")
+        for violation in xml_violations_buffer:
+            logging.debug(violation)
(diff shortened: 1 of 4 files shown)


connector-packager/tests/test_xsd_validator.py
```diff
@@ -65,3 +65,21 @@ def test_validate_single_file(self):
         logging.debug("test_validate_single_file xml violations:")
         for violation in xml_violations_buffer:
             logging.debug(violation)
+
+    def test_validate_vendor_prefix(self):
+
+        test_file = TEST_FOLDER / Path("modular_dialog_connector/connectionFields.xml")
```
lukewrites, line 71: I think part of the magic of `Path` objects is that you can just add a string to the end and they'll make it part of the path.

```suggestion
        test_file = TEST_FOLDER / "modular_dialog_connector/connectionFields.xml"
```

can do the same below.

connector-packager/tests/test_xsd_validator.py
```diff
@@ -65,3 +65,21 @@ def test_validate_single_file(self):
...
+        file_to_test = ConnectorFile("connectionFields.xml", "connection-fields")
+        xml_violations_buffer = []
+
+        self.assertTrue(validate_single_file(file_to_test, test_file, xml_violations_buffer),
+                        "Valid XML file not marked as valid")
+
+        print("\nTest malformed xml. Throws XML validation error.")
+        test_file = TEST_FOLDER / Path("broken_xml/connectionFields.xml")
+        self.assertFalse(validate_single_file(file_to_test, test_file, xml_violations_buffer),
```
lukewrites, line 80: Could we use `[assertRaises](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises)` here instead?

rosswbrown: This is the pattern used by the other tests for connector-packager.  I'll make a note to update all at once later.

connector-packager/tests/test_xsd_validator.py
```diff
@@ -65,3 +65,21 @@ def test_validate_single_file(self):
...
+        file_to_test = ConnectorFile("connectionFields.xml", "connection-fields")
+        xml_violations_buffer = []
+
+        self.assertTrue(validate_single_file(file_to_test, test_file, xml_violations_buffer),
+                        "Valid XML file not marked as valid")
+
+        print("\nTest malformed xml. Throws XML validation error.")
+        test_file = TEST_FOLDER / "broken_xml/connectionFields.xml"
+        self.assertFalse(validate_single_file(file_to_test, test_file, xml_violations_buffer),
+                         "XML file that doesn't follow valid name values marked as valid")
```
pvanderknyff, line 81: This sentence is a bit unclear, took me a few tries to figure out what it was trying to say. How about:

```suggestion
                         "XML file with invalid name values marked as valid")
```

This is a bit more explicit on what has failed when skimming the results.

PR #495, opened 2020-04-08
Truncate long process output results

pvanderknyff:
Had a tdvt results csv file that became very large since the process output was very large (~20,000 characters). Wrote a check that truncates to 512 characters instead. Open to ideas of what the cutoff point should be, it may be a bit restrictive as of now.


pvanderknyff:
Completely forgot to hit merge on this, and now there are merge conflicts. Opened new PR for this change: https://github.com/tableau/connector-plugin-sdk/pull/741

PR #517, opened 2020-05-26
Update server instructions

pvanderknyff:
Tested this a bit ago. Doesn't require messing about with tsm options.

### DIFF
--- a/docs/run-taco.md
+++ b/docs/run-taco.md
@@ -11,14 +11,14 @@ Starting in 2019.4 Beta 1, you can load packaged connectors (otherwise known as 
 ## Run a packaged connector in Tableau Server
 ### Option 1
 For each server node:
-1. Drop your `.taco` file into [Your Tableau Server Install Directory]/data/tabsvc/vizqlserver/Connectors
+1. Drop your `.taco` file into [Your Tableau Server Install Directory]/data/tabsvc/vizqlserver/Connectors. On a default install, this will be in the ProgramData folder. For example:
+`C:\ProgramData\Tableau\Tableau Server\data\tabsvc\vizqlserver\Connectors`
 
 
 ### Option 2
-For each server node:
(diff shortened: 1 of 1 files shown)


joe-langford:
On Windows, can a "server node" have multiple drives, say c: and d:? Reason I ask is that for dev plugins, anyway, my understanding (I got tripped up by it) is that the connector directory has to be on the same drive as the server install. Not sure this would trip anyone up here, but thought you might want to double-check wording of option 2.

pvanderknyff:
> On Windows, can a "server node" have multiple drives, say c: and d:? Reason I ask is that for dev plugins, anyway, my understanding (I got tripped up by it) is that the connector directory has to be on the same drive as the server install. Not sure this would trip anyone up here, but thought you might want to double-check wording of option 2.

I have very little knowledge of how server nodes work, I thought it was a 1-to-1 node-machine relationship. I should probably say "machine" and not "node" in the second option, but in reality someone should go through this process and double check how connector plugins work on nodes.

rosswbrown:
Does this mean they end up in the same directory as partner native connectors? Or are those in a different location after install?

pvanderknyff:
> Does this mean they end up in the same directory as partner native connectors? Or are those in a different location after install?

The system plugins are somewhere else. This is the "repository" for vizqlserver, it's the same folder structure as "My Tableau Repository".

docs/run-taco.md
```diff
@@ -9,9 +9,16 @@ Starting in 2019.4 Beta 1, you can load packaged connectors (otherwise known as
 1. Launch Tableau Desktop.
 
 ## Run a packaged connector in Tableau Server
-1. Create a directory for Tableau connectors. For example:   
+### Option 1
```
rosswbrown, line 12: Option 1 I think is likely the preferred path.  Option 2 is basically the developer and non-packaged, but seems good to document both in the SDK at least.

Issue #505, opened 2020-05-05
DATEPART not applying in .TDD


sablastik:
**About You:**
Name: Konstantin
Company: Citymobil

**Your question:**
I'm trying to customize dialect file and with usual functions like 'YEAR' it works, however when I try to modify date-functions like DATEPART it takes no effect. For example I'm trying to put datetime field to the Text section. Calculation looks like this: "DATEPART('year', [OrderedDate])" and DATEPART is underlined red with this error: "The function 'DATEPART' is not available on non-aggregates for this datasource".

Could you please help me to understand what am I doing wrong?

My TDD looks like this:

```
<?xml version="1.0" encoding="utf-8"?>
<dialect name='test_connector'
         class='test_connector'
         version='18.1'>


  <function-map>

    <function group='string' name='LEFT' return-type='str'>
      <formula>SUBSTR(%1, 1, %2)</formula>
      <argument type='str' />
      <argument type='real' />
    </function>

    <function group='string' name='SPLIT' return-type='str'>
      <formula>SplitByString(%2, %1)[%3]</formula>
      <argument type='str' />
      <argument type='localstr' />
      <argument type='localint' />
    </function>

    <function group='date' name='NOW' return-type='datetime'>
      <formula>NOW()</formula>
    </function>

    <function group='aggregate' name='MAX' return-type='datetime'>
      <formula>MAX(%1)</formula>
      <unagg-formula>%1</unagg-formula>
      <argument type='datetime' />
    </function>

    <function group='date' name='YEAR' return-type='int'>
      <formula>toYear(%1)</formula>
      <argument type='datetime' />
    </function>
    <function group='date' name='YEAR' return-type='int'>
      <formula>toYear(%1)</formula>
      <argument type='date' />
    </function>

    <function group='aggregate' name='MIN' return-type='int'>
      <formula>MIN(%1)</formula>
      <unagg-formula>%1</unagg-formula>
      <argument type='int' />
    </function>

    <date-function name='DATEPART' return-type='int'>  
      <formula part='year'>toYear(%2)</formula>  
      <argument type='localstr' />  
      <argument type='datetime' />  
      <argument type='localstr' />  
    </date-function>  
  </function-map>
  <supported-aggregations>
    <aggregation value='AGG_COUNT'/>
    <aggregation value='AGG_COUNTD'/>
    <aggregation value='AGG_SUM'/>
    <aggregation value='AGG_AVG'/>
    <aggregation value='AGG_MIN'/>
[...]

PR #519, opened 2020-05-28
TFSID 888124 :[SDK] Update custom start of week date functions samples

saugatt:
Remove all except the custom start of the week for date functions with `localstr` as the last argument. See https://github.com/tableau/connector-plugin-sdk/issues/505 for details

### DIFF
--- a/samples/components/dialects/Annotated.tdd
+++ b/samples/components/dialects/Annotated.tdd
@@ -980,7 +980,6 @@
       <argument type='datetime' />
     </date-function>
     <date-function name='DATEDIFF' return-type='int'>
-      <formula>(TO_NUMBER(TO_CHAR(%3,&apos;J&apos;)) - TO_NUMBER(TO_CHAR(%2,&apos;J&apos;)))</formula>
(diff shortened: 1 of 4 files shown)


rosswbrown:
In Annotated.tdd update the formula's for the custom SOW ones to use <formula part='week'> instead of <formula>, [example](https://github.com/tableau/connector-plugin-sdk/blob/master/samples/components/dialects/Annotated.tdd#L996).  It would work as is today I think, but this will make it more precise.

samples/components/dialects/Annotated.tdd
```diff
@@ -1006,7 +1008,8 @@
       <argument type='datetime' />
     </date-function>
     <date-function name='DATEPART' return-type='int'>
-      <formula>FLOOR((7 + TO_NUMBER(TO_CHAR(%2,'DDD')) - 1 + MOD(7 + MOD(TO_NUMBER(TO_CHAR(TRUNC(CAST(%2 AS DATE), 'YYYY'), 'J')) + 1, 7) - %3, 7)) / 7)</formula>
+      <formula>TO_NUMBER(TO_CHAR(%2,&apos;%1&apos; %3&apos;))</formula>
```
rosswbrown, line 1011: remove `<formula>TO_NUMBER(TO_CHAR(%2,&apos;%1&apos; %3&apos;))</formula>`, only need `<formula part='week'> ...` for custom SOW paths.

Same comment for others in this file and full_dialect.tdd

samples/components/dialects/full_dialect.tdd
```diff
@@ -1629,67 +1629,31 @@
     </date-function>
     <date-function name='DATEDIFF' return-type='int'>
       <formula>DATEDIFF(%1,%2,%3)</formula>
```
rosswbrown, line 1631: Needs same change as Annotated.tdd

saugatt: Added the changes

PR #521, opened 2020-05-29
[2020.3] TFSID 1105235: [DUID][DOCS]:  Add delegation documentation to the SDK

devanshsoni9:
**Things To Remember:** 

* **`dev-2020.1`**
* **`2020.2`**
=====


docs/auth-modes.md
```diff
@@ -228,6 +228,42 @@ Note: the ```value``` attribute value for all options is customizable by connect
 })
 ```
 
+### Database impersonation using embedded credentials (DelegationUID)
+
+This applies to databases which support a user that can delegate requests from other user(s). Databases support this by enabling the client to pass a DelegationUID to the server.
+In the case of Tableau server, you can pass the identity of the logged-in user on Tableau Server as the DelegationUID. Tableau would then pass that property to the database using driver properties and authentication is taken care of by the database. This community [article](https://community.tableau.com/docs/DOC-11137) gives more information on Database Impersonation using Embedded credentials.   
+
+For this to work on Tableau Server, user should select the option "Impersonate using embedded password" while publishing the workbook from  Tableau Desktop. Below is an example for how to pass DelegationUID in a JDBC plugin. Please note that some JDBC drivers may have a different name for this property so take a look at driver documentation for the appropriate property name.
```
pvanderknyff, line 236: Do we have an ODBC example? (Does it even make sense with ODBC?)

docs/auth-modes.md
```diff
@@ -228,6 +228,42 @@ Note: the ```value``` attribute value for all options is customizable by connect
 })
 ```
 
+### Database impersonation using embedded credentials (DelegationUID)
+
+This applies to databases which support a user that can delegate requests from other user(s). Databases support this by enabling the client to pass a DelegationUID to the server.
+In the case of Tableau server, you can pass the identity of the logged-in user on Tableau Server as the DelegationUID. Tableau would then pass that property to the database using driver properties and authentication is taken care of by the database. This community [article](https://community.tableau.com/docs/DOC-11137) gives more information on Database Impersonation using Embedded credentials.
```
pvanderknyff, line 234: Does it work/make sense on desktop? Do you have to do extra work in the connector files to use it on server?

docs/auth-modes.md
```diff
@@ -228,6 +228,42 @@ Note: the ```value``` attribute value for all options is customizable by connect
...
+ 
+ Also add CAP_AUTH_DB_IMPERSONATE to the list of capabilities in plugin manifest so that the option "Impersonate using embedded password" shows up in the publish dialog.
+ ```xml 
+   // manifest.xml 
+    <customizations>
+      <customization name="CAP_AUTH_DB_IMPERSONATE" value="yes"/>
+    </customizations>
+ ```
+ 
+ Impala delegation sample link(TBA)
```
pvanderknyff, line 265: Don't forget to update this

docs/auth-modes.md
```diff
@@ -228,6 +228,42 @@ Note: the ```value``` attribute value for all options is customizable by connect
...
+
+    var props = {};
+    props["UID"] = attr[connectionHelper.attributeUsername];
+    props["PWD"] = attr[connectionHelper.attributePassword];
+   
+    if (attr[connectionHelper.attributeTableauServerAuthMode] == connectionHelper.valueAuthModeDBImpersonate) {
+        var str = attr[connectionHelper.attributeTableauServerUser];
+        
+        if (!isEmpty(str)){
+            props["DelegationUID"] = str;
```
pvanderknyff, line 252: Does this need to be added to the list of required attributes in the connection resolver?

devanshsoni9:
Hey @pvanderknyff 
I do not see gh-pages-dev-2020.3 branch. It seems like it needs to be created. I can create one if you want me to

docs/auth-modes.md
devanshsoni9: yeah, it should work fine with ODBC. 
I haven't gotten the chance to test with ODBC.

docs/auth-modes.md
devanshsoni9: Delegation is useful on Tableau Server. We do not support it on Desktop. On Desktop the user would just login using their username-password. 
The changes in connectionProperties and manifest are required to be present at the server level, however desktop plugin can be just a simple jdbc plugin supporting Basic Auth

docs/auth-modes.md
devanshsoni9: sent another PR for the sample -  https://github.com/tableau/connector-plugin-sdk/pull/522

docs/auth-modes.md
devanshsoni9: I don't think so. They just need to be passed to the propertiesbuilder and read once. We don't need to persist them. SetImpersonateCurrentUser() is called before resolver->*builder is called which ensures that the tableauServerUser is set.

pvanderknyff:
> Hey @pvanderknyff
> I do not see gh-pages-dev-2020.3 branch. It seems like it needs to be created. I can create one if you want me to

Sorry, it wasn't created. Just made it, you'll need to close this PR and open a new one.

docs/auth-modes.md
pvanderknyff: May want to re-word the description to say that explicitly, the phrase "in the case of Tableau Server" implies there is a Desktop case.

docs/auth-modes.md
devanshsoni9: I am thinking the link would be something in the tree of master branch, and not dev-2020-3 branch. So I guess I should link to this path https://github.com/tableau/connector-plugin-sdk/tree/master/samples/plugins/db_impersonation which does not exist yet, but will exist once the merge from 2020-3 to master takes place

docs/api-reference.md
```diff
@@ -155,22 +155,30 @@ Use care when logging so that you don’t expose sensitive information like pass
...
+| attributeSSLCert               | Connection attribute for the SSL Certfile                   |
+| attributeSSLMode               | Connection attribute for the SSL Mode                       |
+| attributeTableauServerAuthMode | Connection attribute for tableau server authentication mode |
+| attributeTableauServerUser     | Connection attribute for tableau server user                |
+| attributeUsername              | Connection attribute for the user name                      |
+| attributeWarehouse             | Connection attribute for the Warehouse                      |
+| keywordODBCUsername            | ODBC Username keyword                                       |
+| keywordODBCPassword            | ODBC Password keyword                                       |
+
+_Attribute values_
```
pvanderknyff, line 177: I'm not sure we need a separate table for this. These are more string constants, and valueAuthModeDBImpersoante fits that well enough.

rosswbrown:
Before release let's follow the example of Password Only with a note about when support was added.

PR #522, opened 2020-05-29
[2020.3] Impala Plugin Sample  - Database impersonation using embedded credentials (DelegationUID)

devanshsoni9:
This is provided as a sample to the documentation PR https://github.com/tableau/connector-plugin-sdk/pull/521
Pending Internal testing for the dialect since I have changed the base dialect since first version of this CR.

**Things To Remember:** 

* **`dev-2020.1`**
* **`2020.2`**
=====


samples/plugins/impala_jdbc_delegationUID/dialect.tdd
```diff
@@ -0,0 +1,12 @@
...
+         class='myimpala'
+         base='PostgreSQL90Dialect'
+         version='18.1'>
+   <function-map>
+    <function name='FLOAT' group='cast' return-type='real'>
+      <formula>CAST(((EXTRACT(EPOCH FROM %1) / ( 60 * 60 * 24)) + (365 * 70 + 17)) AS BIGINT)</formula>
+      <argument type='date' />
+    </function>
+   </function-map>
+</dialect>
```
pvanderknyff, line 12: This is the same as the postgres dialect, which is incompatible with the Impala dialect. You can use the impala dialect as a base I think, we also have a hive example tdd if needed.

devanshsoni9: I modified base dialect to Impala23Dialect however I cannot test this since I don't have a test environment after losing access. I tried to create a test environment on public AWS but the cloudera [quickstart](https://docs.aws.amazon.com/quickstart/latest/cloudera/welcome.html) cloudformation scripts failed to create CFN stack at first. However when I setup my own VPC, the scripts did pass and I was able to create a director instance. However the cfn scripts did not download and setup other config needed for creating nodes and I figured it would require a lot of manual labor to setup properly. I suggest we should test it in internal test env with the new base dialect if this is one of the release criteria.

devanshsoni9:
> Does this also require 2020.3? If so, you need to merge into dev-2020.3, which I just created.

changed the base branch to dev-2020.3

samples/plugins/impala_jdbc_delegationUID/connectionProperties.js
```diff
@@ -0,0 +1,24 @@
...
+    function isEmpty(str) {
+        return (!str || 0 === str.length); 
+    }
+
+    var props = {};
+    props["UID"] = attr[connectionHelper.attributeUsername];
+    props["PWD"] = attr[connectionHelper.attributePassword];
+   
+    if (attr[connectionHelper.attributeTableauServerAuthMode] == connectionHelper.valueAuthModeDBImpersonate) {
+        var str = attr[connectionHelper.attributeTableauServerUser];
```
devanshsoni9, line 12: We should be adding these new constants to the SDK docs as well.

samples/plugins/impala_jdbc_delegationUID/dialect.tdd
pvanderknyff: It doesn't have to be a fully-fledged connector to fulfill the role of demonstrating database impersonation. However, some people may assume that it is an exmaple of an Impala connector, and has the same functionality as our built-in one. They will think can make an impala-based connector using this as a starting point. 

If it's possible to use postgres instead to demonstrate DelegationUID, then that might be a better option. It may also be a good idea to rename the sample to something like "database_impersonation_example", so that people know that the purpose of the connector is to demonstrate database impersonation and not how to make an impala-based connector.

devanshsoni9: I am afraid, I think postgres does not support this kind of database impersonation using delegation UID or at-least I do not know of that right now. Postgres does support kerberos delegation, however these changes are not about that and we plan to support those later.

I will rename the sample plugin

devanshsoni9:
thanks @pvanderknyff. I will merge this one. If you have time, can you also take a look at the documentation one.

devanshsoni9:
@pvanderknyff when do we merge these changes into master ? Do we do a bulk merge into master after the release?

pvanderknyff:
Yeah, they get bulk merged once 2020.3 is released.

samples/plugins/db_impersonation/connectionBuilder.js
```diff
@@ -0,0 +1,5 @@
+(function dsbuilder(attr) {
+    var urlBuilder = "jdbc:impala://" + attr[connectionHelper.attributeServer] + ":" + attr[connectionHelper.attributePort] + ";" + "AuthMech=3;";
```
rosswbrown, line 2: We should add a comment to the driver name/version this sample was created with

samples/plugins/db_impersonation/connectionProperties.js
```diff
@@ -0,0 +1,24 @@
...
+
+    var props = {};
+    props["UID"] = attr[connectionHelper.attributeUsername];
+    props["PWD"] = attr[connectionHelper.attributePassword];
+   
+    if (attr[connectionHelper.attributeTableauServerAuthMode] == connectionHelper.valueAuthModeDBImpersonate) {
+        var str = attr[connectionHelper.attributeTableauServerUser];
+        
+        if (!isEmpty(str)){
+            props["DelegationUID"] = str;
```
rosswbrown, line 15: Add a comment to the doc linked in gh-pages as well: https://impala.apache.org/docs/build/html/topics/impala_delegation.html

samples/plugins/db_impersonation/manifest.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version='1.0' encoding='utf-8' ?>
+
+<connector-plugin class='impala_jdbc_delegationUID' superclass='jdbc' plugin-version='1.0.0' name='Impala JDBC Plugin' version='19.4'>
+  <vendor-information>
+    <company name="Company"/>
+    <support-link url = "https://github.com/tableau/connector-plugin-sdk"/>
+    <driver-download-link url = "some url"/>
```
rosswbrown, line 7: We should match other samples:
```
      <company name="Company Name"/>
      <support-link url="https://example.com"/>
      <driver-download-link url="https://drivers.example.com"/>
```

samples/plugins/db_impersonation/connection-dialog.tcd
```diff
@@ -0,0 +1,13 @@
+<connection-dialog class='impala_jdbc_delegationUID'>
+  <connection-config>
+    <authentication-mode value='Basic' />
+    <authentication-options>
+      <option name="UsernameAndPassword" default="true" />
+    </authentication-options>
+    <has-databases value="true" />
+    <has-schemas value="true" />
+    <port-prompt value="Port: " default="21050" />
+    <show-db-browse-button value="true" />
```
rosswbrown, line 10: Remove `<show-db-browse-button value="true" />`

samples/plugins/db_impersonation/connectionProperties.js
```diff
@@ -0,0 +1,24 @@
...
+        }
+    }
+
+    if (attr["sslmode"] !== "") {
+        props["ssl"] = "true";
```
rosswbrown, line 20: Impala documentation shows "SSL" and "0" or "1" as key/value for SSL.

Issue #518, opened 2020-05-27
[Doc BUG] Connection resolver language on example page is confusing


sidwray:
**Describe the bug**
This line on the Examples page is a little confusing and is leading a few folks to believe that they could make an OData plugin that isn't ODBC or JDBC based. 

https://tableau.github.io/connector-plugin-sdk/docs/example

Tableau Connection Resolver file creates the connection to your file using JavaScript files to define and make the connection. **_(If the base is not set to “odbc” or “jdbc”, then this is file may not be required.)_**

jainam1995:
The fix has been merged. Closing.

PR #523, opened 2020-06-01
Fix some wording on the example page regarding the .tcd and .tdr files

jainam1995:
Issue : https://github.com/tableau/connector-plugin-sdk/issues/518 .
Tried rewording the statements to keep the intent but make it less confusing. Open to better re structuring of the statements.

### DIFF
--- a/docs/example.md
+++ b/docs/example.md
@@ -6,8 +6,8 @@ A Tableau connector is a set of files that describe the UI elements needed to co
 
 Tableau connector files include:
 - __Manifest file__ that tells Tableau about the connector. Tableau uses the name from this file to add it to the list of available connectors in the UI.
-- __Tableau Custom Dialog file__ customizes the connection dialog as needed for this connection. (If the base is not set to "odbc" or "jdbc", then this is file may not be required.)
-- __Tableau Connection Resolver file__ creates the connection to your file using JavaScript files to define and make the connection. (If the base is not set to "odbc" or "jdbc", then this is file may not be required.)
+- __Tableau Custom Dialog file__ customizes the connection dialog as needed for this connection. This file is optional. If not created, it will be based off the superclass set in the Manifest file (manifest.xml) by default. However if the superclass is set to "odbc" or "jdbc", then this file is required.
+- __Tableau Connection Resolver file__ creates the connection to your file using JavaScript files to define and make the connection. This file is optional. If not created, it will be based off the superclass set in the Manifest file (manifest.xml) by default. However if the superclass is set to "odbc" or "jdbc", then this file is required.
 - __Tableau Dialect file__ identifies which SQL dialect to use.
 
 Tableau provides a base set of connector files that you use to create your own customized connector. You can use the connectivity test harness to validate the connector behavior as you build it. 
--- a/docs/example.md
+++ b/docs/example.md
@@ -6,8 +6,8 @@ A Tableau connector is a set of files that describe the UI elements needed to co
 
 Tableau connector files include:
 - __Manifest file__ that tells Tableau about the connector. Tableau uses the name from this file to add it to the list of available connectors in the UI.
-- __Tableau Custom Dialog file__ customizes the connection dialog as needed for this connection. This file is optional. If not created, it will be based off the superclass set in the Manifest file (manifest.xml) by default. However if the superclass is set to "odbc" or "jdbc", then this file is required.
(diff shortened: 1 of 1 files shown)


docs/example.md
```diff
@@ -6,8 +6,8 @@ A Tableau connector is a set of files that describe the UI elements needed to co
...
-- __Tableau Custom Dialog file__ customizes the connection dialog as needed for this connection. (If the base is not set to "odbc" or "jdbc", then this is file may not be required.)
-- __Tableau Connection Resolver file__ creates the connection to your file using JavaScript files to define and make the connection. (If the base is not set to "odbc" or "jdbc", then this is file may not be required.)
+- __Tableau Custom Dialog file__ customizes the connection dialog as needed for this connection. This file is optional. If not created, it will be based off the superclass set in the Manifest file (manifest.xml) by default. However if the superclass is set to "odbc" or "jdbc", then this file is required.
+- __Tableau Connection Resolver file__ creates the connection to your file using JavaScript files to define and make the connection. This file is optional. If not created, it will be based off the superclass set in the Manifest file (manifest.xml) by default. However if the superclass is set to "odbc" or "jdbc", then this file is required.
```
pvanderknyff, line 10: I think this is written backwards. 99% of people should be using odbc or jdbc, so in 99% of the people reading this will be required to have these files. However, as it stands the first thing they read is "This file is optional," full stop, and they may not read further. 

I actually think the original is clearer on this - this file _may not_ be required in certain, very specific situations. I would hesitate to change that narrative.

pvanderknyff:
The original issue is about people mistaking the line to mean they can do something that isn't odbc or jdbc based, this leaves that ambiguity.

jainam1995:
> The original issue is about people mistaking the line to mean they can do something that isn't odbc or jdbc based, this leaves that ambiguity.

Yes, the options other than odbc or jdbc are the classes they mention in the superclass . If these are not odbc and jdbc , in that case this file becomes optional but if they mention it to be odbc or jdbc then this file is required. That's what i tried to point out by mentioning the superclass information in the change. 

I will rephrase the change to not change the "may to be required".

docs/example.md
jainam1995: I will rephrase it to keep the original narrative,

jainam1995:
> The original issue is about people mistaking the line to mean they can do something that isn't odbc or jdbc based, this leaves that ambiguity.

Does adding information about the usperclass which is the throd and only option after odbc and jdbc help in the direction of solving the ambiguity ?

docs/example.md
jainam1995: Was thinking of something like  (If the superclass in Manifest file (manifest.xml) is not set to the value “odbc” or “jdbc”, then this is file may not be required and it will be based off the odbc/jdbc based superclass specified.

docs/example.md
```diff
@@ -6,8 +6,8 @@ A Tableau connector is a set of files that describe the UI elements needed to co
 
 Tableau connector files include:
 - __Manifest file__ that tells Tableau about the connector. Tableau uses the name from this file to add it to the list of available connectors in the UI.
-- __Tableau Custom Dialog file__ customizes the connection dialog as needed for this connection. (If the base is not set to "odbc" or "jdbc", then this is file may not be required.)
-- __Tableau Connection Resolver file__ creates the connection to your file using JavaScript files to define and make the connection. (If the base is not set to "odbc" or "jdbc", then this is file may not be required.)
+- __Tableau Custom Dialog file__ customizes the connection dialog as needed for this connection. This file is optional. If not created, it will be based off the superclass set in the Manifest file (manifest.xml) by default. However if the superclass is set to "odbc" or "jdbc", then this file is required.
```
pvanderknyff, line 9:
```suggestion
- __Tableau Custom Dialog file__ customizes the connection dialog as needed for this connection. In most cases, this is required. However, if you subclass from an existing connector like MySQL or Postgres, and you do not create this file, we will use the default for that connector.
```

pvanderknyff, line 9:
```suggestion
- __Tableau Custom Dialog file__ customizes the connection dialog as needed for this connection.
```

pvanderknyff: This would be my preferred option, since it's unlikely and discouraged for them to a) use an existing connector as a superclass and b) if they do use an existing connector, use the base files.

jainam1995: This is what i prefer too after looking at the confusion the term optional is creating. I made the necessary changes,

PR #526, opened 2020-06-02
Update dialect XSD for 'format-true' and 'format-false' to allow literal, predicate and value attributes

rosswbrown:
'format-true' and 'format-false' are used in two use cases: literals and predicates

Literals: when a boolean is used a value
  - defaults to x ? 1 : 0
  - example usage: SELECT 1 as literalBool

Predicates: when a boolean is used as a predicate
  - defaults to isTrue ? (1=1) : (1=0)
  - example usage: SELECT something as blah WHERE (1=1)
  - 'value' attribute is used as 'predicate' for backwards compatability


rosswbrown:
> Did you mean to have the samples updated in this pull request as well?

I couldn't decide if it made sense to do them together or not.  I thought if someone was looking in change history the comments might be easier to follow if they're apart.

PR #531, opened 2020-06-10
Add SQLDialect to dialect examples

jsun28:
Add SQLDialect to dialect examples to capture all implementation in BaseDialect\SQLDialect.
Manual tested tableau desktop can load this tdd, no syntax errors.

Methods in SQLDialect that can not be captured:
1. Methods are not part of dialect, these methods serve as a helper method for individual dialect to use.
    AddNativeSplitFunction
    AddRecursiveSplitFunction
    GetIfThenElseFormula
    NoCaseAliasLessThan
    BinaryAliasLessThan
    TruncateRemoteAlias
    FormatInsertParameterMarker
    GetParameterMarkers
    FormatBulkInsert
    AddFunctionRaw
    HexEncodeNonAscii
    ReplaceInvalidAscii

2. TDD doesn't support those functions now, and these methods are only overridden by a few dialects.
    GetAliasLessThanFunc
    FormatInitialSQL 
    FormatInsertDataValue DataValueFormatter is partially covered in tdd
    FormatUnion
    FormatRange
    FormatExecStoredProcInto
    FormatTableDEE
    FormatSelectMember
    FormatExplainPlan
    FormatColumnDefinition's real type, because it requires column's precision, see SQLDialect's FormatColumnDefinition for details
    FormatColumnDefinition
    FormatCreateSchema
    FormatDropSchema    
    FormatFromClause
    FormatForeignKey
    FormatPrimaryKey
    FormatUniqueConstraint
    FormatAddPersistedComputedColumn
    FormatDropColumn
    FormatPlaceholder
    FormatIsNull
    FormatIsNotNull
    FormatExecuteAsUser
    FormatRemoteQuery
    FormatCreateDatabase
    FormatDropDatabase

3. Methods are not supported in tdd and are not overridden by any dialects
    FormatUpdate
    FormatLateralValuesClause
    FormatAppendInto
    FormatDelete
    FormatInsertIntoSelect
    FormatWindowFunction
    FormatFrame
    FormatPreparedStatement
    FormatExecuteStatement
    FormatTaggedQueryString

### DIFF
--- a/samples/components/dialects/SQLDialect.tdd
+++ b/samples/components/dialects/SQLDialect.tdd
@@ -1,16 +1,15 @@
 <?xml version='1.0' encoding='utf-8' ?>
 
 <!--
-  This dialect is the implementation of SQLDialect and BaseDialect.
+  This is the base implementation of SQL dialects.
   dialect
   Name: The name of the dialect. Can be referenced as the base of another dialect for inheritance.
   Class: The class of the connector the dialect is associated with
-  Dialect-version (optional): The minimum database version this dialect should be used with
   Base (optional): The name of the dialect that this one builds off of.
                    All functions and properties default to the base dialect's implementation.
   Version: Tableau document version
 -->
-<dialect class='example' dialect-version='1.0' name='SQLDialect' version='18.1'>
+<dialect class='example' name='SQLDialect' version='18.1'>
   <!--
     function-map
     Map of Tableau functions to SQL expressions. Can include function, date-function,
@@ -178,11 +177,12 @@
       where no type-specific formula is given.
     -->
     <format-null>
-      <local-type name='real' value='NULL' />
       <local-type name='int' value='NULL' />
+      <local-type name='real' value='NULL' />
+      <local-type name='str' value='NULL' />
+      <local-type name='datetime' value='NULL' />
       <local-type name='bool' value='NULL' />
       <local-type name='date' value='NULL' />
-      <local-type name='datetime' value='NULL' />
       <local-type name='localint' value='NULL' />
       <local-type name='localreal' value='NULL' />
       <local-type name='localstr' value='NULL' />
--- a/samples/components/dialects/SQLDialect.tdd
+++ b/samples/components/dialects/SQLDialect.tdd
@@ -183,9 +183,6 @@
       <local-type name='datetime' value='NULL' />
       <local-type name='bool' value='NULL' />
       <local-type name='date' value='NULL' />
-      <local-type name='localint' value='NULL' />
-      <local-type name='localreal' value='NULL' />
-      <local-type name='localstr' value='NULL' />
       <local-type name='spatial' value='NULL' />
     </format-null>
     <!--
--- a/samples/components/dialects/SQLDialect.tdd
+++ b/samples/components/dialects/SQLDialect.tdd
@@ -1,0 +1,273 @@
+<?xml version='1.0' encoding='utf-8' ?>
+
+<!--
+  This dialect is the implementation of SQLDialect and BaseDialect.
+  dialect
+  Name: The name of the dialect. Can be referenced as the base of another dialect for inheritance.
+  Class: The class of the connector the dialect is associated with
+  Dialect-version (optional): The minimum database version this dialect should be used with
+  Base (optional): The name of the dialect that this one builds off of.
+                   All functions and properties default to the base dialect's implementation.
+  Version: Tableau document version
+-->
+<dialect class='example' dialect-version='1.0' name='SQLDialect' version='18.1'>
+  <!--
+    function-map
+    Map of Tableau functions to SQL expressions. Can include function, date-function,
+    remove-function (rare), native-split-function, and recursive-split-function.
+  -->
+  <function-map>
+    <!--
+      function
+      Group: Tableau function group. Multiple groups are comma-separated.
+      Name: Function name
+      Return-type: Tableau data type. These include bool, date, datetime,
+                    int, real, spatial, str
+    -->
+    <function group='numeric' name='DIV' return-type='int'>
+      <!--
+        formula
+        SQL expression formula. Input parameters are denoted with %1, %2, etc.
+      -->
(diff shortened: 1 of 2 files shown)


samples/components/dialects/SQLDialect.tdd
```diff
@@ -0,0 +1,273 @@
+<?xml version='1.0' encoding='utf-8' ?>
+
+<!--
+  This dialect is the implementation of SQLDialect and BaseDialect.
```
pvanderknyff, line 4: Does this implement BaseDialect? It looks like it's just for SQLdialect.

jsun28: yes, it implemented BaseDialect as well, actually only FormatSetIsolationLevel and FormatStartOfWeek are from basedialect.
Because there is not too much dialect implementation in basedialect, the majority of methods there are some dialect internal functions or helper functions to be used by individual dialects.

jsun28: SQLDialect is an abstract class, no datasource can use it directly, it has to be used as a base and subclass from it. I don't think it can pass tdvt test. I think the idea of this is to provide details of our base implementation of sqldialect/basedialect, so partners can understand it well while developing their own tdd.

samples/components/dialects/SQLDialect.tdd
```diff
@@ -0,0 +1,273 @@
...
+      Allows a type-specific expression for NULL. Tableau defaults to 'NULL'
+      where no type-specific formula is given.
+    -->
+    <format-null>
+      <local-type name='real' value='NULL' />
+      <local-type name='int' value='NULL' />
+      <local-type name='bool' value='NULL' />
+      <local-type name='date' value='NULL' />
+      <local-type name='datetime' value='NULL' />
+      <local-type name='localint' value='NULL' />
```
rosswbrown, line 186: local-type should be from the internal DataType enum (codegen/coredata/enums.data)

samples/components/dialects/SQLDialect.tdd
```diff
@@ -0,0 +1,273 @@
...
+      format-set-isolation-level
+      Formula for setting session isolation level
+    -->
+    <format-set-isolation-level value='SET TRANSACTION ISOLATION LEVEL %1' />
+    <!--
+      format-simple-case
+      [Default]Case
+      BalancedIIF
+    -->
+    <format-simple-case value='Case' />
```
rosswbrown, line 224: I think same comment as format-if-then-else, double check me

samples/components/dialects/SQLDialect.tdd
```diff
@@ -0,0 +1,273 @@
+<?xml version='1.0' encoding='utf-8' ?>
+
+<!--
+  This dialect is the implementation of SQLDialect and BaseDialect.
+  dialect
+  Name: The name of the dialect. Can be referenced as the base of another dialect for inheritance.
+  Class: The class of the connector the dialect is associated with
+  Dialect-version (optional): The minimum database version this dialect should be used with
```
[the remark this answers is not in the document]
jsun28: Looked at tdd wiki, https://mytableau.tableaucorp.com/pages/viewpage.action?pageId=126288933, Dialect-version should match server instance version, so should be good in plugin scenario, so removed it.

samples/components/dialects/SQLDialect.tdd
jsun28: Based on my manual testing, local-type is restricted within those values https://github.com/tableau/connector-plugin-sdk/blob/master/validation/tdd_latest.xsd#L102
If a type not in this list, tdd won't be loaded.

samples/components/dialects/SQLDialect.tdd
```diff
@@ -0,0 +1,273 @@
...
+    <!--
+      format-false
+      String literal or expression for boolean false
+    -->
+    <format-false value='(1=0)' />
+    <!--
+      format-if-then-else
+      Strategy for formatting if/then/else statement
+      [Default]Case: (CASE WHEN %1 THEN %2 ELSE %3 END)
+      IIF: Use IIF formula
```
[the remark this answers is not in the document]
jsun28: yeah, agree, in IIF case, GetIfThenElseFormula should return IIF(%1,%2,%3), but in [datadrivendialect's code](https://opengrok.prod.tableautools.com/source/xref/teams_near/modules/connectors/tabmixins/main/db/DataDrivenDialect.cpp#429), the GetIfThenElseFormula's result is same as Case option.
I'll open a story to support customization of GetIfThenElseFormula.

samples/components/dialects/SQLDialect.tdd
rosswbrown: It looks like in practice only the DataType enum values can ever be used.  See DataDrivenDialect::FormatNULL where Expression::ExprTypeFromDataType is used to go from DataType ->ExprType.  Maybe @erademacher can double check me here.

erademacher: Ross is right. Maybe I should've added a DataType enum to the XSD instead of letting ArgumentType cover both DataType and Expression::Type (which is basically a superset of DataType that includes local types for literals in expressions), but for now, it is what it is.

Local types don't really make sense in this context, so please remove them to avoid confusion. Also, it looks like you're missing 'str'.

jsun28: Removed local types in the latest commit, 'str' is also added.
Talked a bit with Evan, maybe we can create a follow-up story to add DataType enum to the XSD to let format-null use it, what is your opinion? @rosswbrown

rosswbrown: Yeah I think it makes sense to open something up to consider it at least.  It would be a breaking change at the XSD level, but in practice its already a runtime error, so I think there should be limited impact.  It would be 2020.4 or later timeframe either way.

rosswbrown:
Can you add the information in this PR's description into the Description of 963307.  Its a nice summary of where we are at with overall functionality parity.

samples/components/dialects/SQLDialect.tdd
jsun28: created a story as 1130681

PR #545, opened 2020-06-26
Add format-string-literal documentation

pvanderknyff:
Add some documentation for the format-string-literal change I made a while ago.

There's also a bunch of whitespace changes, since I have my VSCode trim trailing spaces when I save I think. Doesn't look like anything actually changed, but the diff is a bit squirrely.


dmoynihan:
looks good to me

PR #551, opened 2020-07-06
Add commands in tdvt to run the connectors test command already added to tabquerycli

jainam1995:
Based off a discussion with luke on the present flow of control for tdvt which heavily relies on BatchWorkQueue to run all the current datasource specific tdvt tests, adding a separate flow for running the connector tests which does not require any of those and only needs two arguments seemed as the cleanest way forward.

This code review is based on that discussion.

Tested by running the tests from tdvt and the results were as expected and the same as that run using tabquerycli directly.

### DIFF
--- a/tdvt/tdvt/tdvt.py
+++ b/tdvt/tdvt/tdvt.py
@@ -417,7 +417,7 @@ run_pattern_usage_text = '''
     '''
 run_connectors_test_usage_text = '''
     The commands below can be used to run the connectors tests.
-    'filename.txt' in the below commands is the path of the xml test file used to run the tests.
+    'filename.xml' in the below commands is the path of the xml test file used to run the tests.
     Run ConnectionBuilderTest
         run-connectors-test --conn-test connectionBuilder --conn-test-file filepath.xml
     Run NormalizeConnectionAttributes Test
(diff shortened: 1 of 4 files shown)


tdvt/tdvt/tdvt.py
```diff
@@ -415,6 +415,21 @@ def get_level_of_parallelization(args):
         run-pattern postgres_odbc --logp logicaltests/setup/calcs/setup.BUGS.*.dbo.xml --tdp cast_calcs.*.tds --test-ex 59740
 
     '''
+run_connectors_test_usage_text = '''
```
lukewrites, line 418: This may be a big nit, but I wonder if this command should start with `test` instead of `run`. The other `run` commands are for the TDVT tool itself.

tdvt/tdvt/tdvt.py
```diff
@@ -415,6 +415,21 @@ def get_level_of_parallelization(args):
...
+    The commands below can be used to run the connectors tests.
+    'filename.txt' in the below commands is the path of the xml test file used to run the tests.
```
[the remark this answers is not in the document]
jainam1995: Will fix

tdvt/tdvt/tdvt.py
```diff
@@ -475,6 +490,11 @@ def create_parser():
     run_file_parser = subparsers.add_parser('run-file', help='Run tests from a file.', parents=[run_test_common_parser], usage=run_file_usage_text)
     run_file_parser.add_argument('run_file', help='Json file containing failed tests to run.')
 
+    #Run Connectors Test
+    run_connectors_test_parser = subparsers.add_parser('run-connectors-test', help='Run a connectors test using a file', parents=[run_test_common_parser], usage=run_connectors_test_usage_text)
+    run_connectors_test_parser.add_argument('--conn-test', dest='conn_test', help='Name of the Connectors Test to run.',  required=True, default=None, const='', nargs='?')
```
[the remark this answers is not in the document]
jainam1995: Removed default=None, const='', nargs='?' from both

tdvt/tdvt/tdvt.py
jainam1995: I had used the name run_connectors_test because we are running the connectors test using this command. Also i may be misunderstanding the second statement, but isnt this command also for the tdvt tool ?

jainam1995:
> Make sure to run the unit tests, too.

Will be adding a unit test for this change in the next revision

tdvt/tdvt/tdvt.py
lukewrites: Sorry, my misunderstanding. Since this is testing the connector, not the TDVT tool, `run` is the correct start :)

jainam1995: Ran the args test which take care of the change because the change is basically the parsing of args

jainam1995:
> Make sure to run the unit tests, too.

ran the tests

PR #553, opened 2020-07-07
Added validation rule enforcing default values for non-optional fields in the advanced category

erademacher:
This kind of requirement isn't really enforceable in XSD, so I added it in the python validator.

### DIFF
--- a/connector-packager/connector_packager/xsd_validator.py
+++ b/connector-packager/connector_packager/xsd_validator.py
@@ -159,4 +159,11 @@ def validate_file_specific_rules(file_to_test: ConnectorFile, path_to_file: Path
                                                  "' not an allowed value. See 'Connection Field Platform Integration' section of documentation for allowed values.")
                     return False
 
+                category = child.attrib['category']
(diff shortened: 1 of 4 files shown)


connector-packager/connector_packager/xsd_validator.py
```diff
@@ -159,4 +159,12 @@ def validate_file_specific_rules(file_to_test: ConnectorFile, path_to_file: Path
                                                  "' not an allowed value. See 'Connection Field Platform Integration' section of documentation for allowed values.")
                     return False
 
+            if 'category' in child.attrib:
+                category = child.attrib['category']
+                optional = 'optional' not in child.attrib or child.attrib['optional'] == 'true'
+                if category == 'advanced' and not optional and 'default-value' not in child.attrib:
```
sxlin, line 165: in m-c-d, the value of empty string is treated as no user input.  So if the default value is an empty string, we will still have problem. 

I propose we simple enforce the fields in Advanced tab must be optional.

PR #554, opened 2020-07-08
Make SQLiteODBCExtract sample dialect independent of any base dialect

jsun28:
Export the functionality of CDataExtractDialec as a TDD, to ensure SQLiteODBCExtract doesn't need to inherit from anything. 

Function map is generated by WriteDialectUnTest, I added all SQL formatting stuff manually based on content in CDataExtractDialect, GenericODBCDetectedDialect, and GenericODBCDialect, manually tested it with some queries in desktop.


samples/scenarios/extract_only/sqlite_extract/dialect.xml
```diff
@@ -1,5 +1,1319 @@
...
+        <remote-type name='default' value='DATE' />
+      </local-type>
+      <local-type name='bool'>
+        <remote-type name='default' value='BOOLEAN' />
+      </local-type>
+      <local-type name='spatial'>
+        <remote-type name='default' value='GEOGRAPHY' />
+      </local-type>
+    </format-column-definition>
+    <format-create-table>
```
rosswbrown, line 1261: Is there any way we can get to a more minimal representation of this dialect that still works (it could be slightly different).
I don't imagine for example format-create-table, format-drop-table, format-set-isolation-level, format-stored-proc-call matter in extract only context.
Others like format-order-by and format-null match the [default](https://github.com/tableau/connector-plugin-sdk/blob/dev-2020.3/samples/components/dialects/SQLDialect.tdd) (it looks like to me anyway), so I think those might be able to be removed as well.

samples/scenarios/extract_only/sqlite_extract/dialect.xml
```diff
@@ -1,5 +1,1319 @@
...
+        <remote-type name='DATE' value='DATE' />
+        <remote-type name='default' value='TIMESTAMP' />
+      </local-type>
+      <local-type name='date'>
+        <remote-type name='default' value='DATE' />
+      </local-type>
+      <local-type name='bool'>
+        <remote-type name='default' value='BOOLEAN' />
+      </local-type>
+      <local-type name='spatial'>
```
rosswbrown, line 1257: I don't know, but I doubt sqlite supports spatial type.

samples/scenarios/extract_only/sqlite_extract/dialect.xml
jsun28: I removed all SQL formatting stuff that are same with SQLDialect in 4th and 5th commit, also including format-create-table, format-drop-table, format-set-isolation-level, format-stored-proc-call.

samples/scenarios/extract_only/sqlite_extract/dialect.xml
jsun28: [Googled some info about this](https://pro.arcgis.com/en/pro-app/help/data/databases/overview-database-spatial-types.htm), it looks like sqlite supports spatial type, I originally added spatial type was because [CDataExtractDialect](https://opengrok.prod.tableautools.com/source/xref/teams_near/modules/connectors/tabmixins/main/db/CDataExtractDialect.cpp#123) has it.
so I removed spatial type, and after removing it, format-column-definition is same with base implementation, so deleted the whole format-column-definition.

samples/scenarios/extract_only/sqlite_extract/dialect.xml
```diff
@@ -1,5 +1,1251 @@
...
+      <formula>CAST(%1 AS FLOAT) ^ %2</formula>
+      <argument type='int' />
+      <argument type='int' />
+    </function>
+    <function group='operator' name='||' return-type='bool'>
+      <formula>(%1 OR %2)</formula>
+      <argument type='bool' />
+      <argument type='bool' />
+    </function>
+    <date-function name='DATEADD' return-type='datetime'>
```
rosswbrown, line 1065: Missed this the first time.  For all of these date-function ones the export tool doesn't output the arguments correctly, they are all off by %1 is always the date-part-group value: https://tableau.github.io/connector-plugin-sdk/docs/dialect.  The 'sunday' in each of these is also erroneous.

Check out my recent conversion review for some working examples or the postgres example.

Its entirely possible some (or possibly all) are not even needed for extract-only, but I don't know that for sure, you might try a manual test.

jsun28: My manual test process: Create a calculation field for each date function, copy and paste some test cases from tdvt, for example `DATETRUNC('day', DATETIME([Datetime0]), 'monday')` for DATETRUNC, the calculation fields' result are all same with tdvt's standard result, also same with the result of using original dialect(CDataExtractDialect as base).

so I think it is ok to remove all date functions, thanks for pointing this out.

PR #556, opened 2020-07-13
TFSID1140126: [SDK] Add password-only authentication type in  validation XSD in SDK

saugatt:
Update tcd_latest.xsd to have PasswordOnly as AuthMode-CT


validation/tcd_latest.xsd
```diff
@@ -50,6 +50,7 @@
       <xs:enumeration value="SharePoint"/>
       <xs:enumeration value="SAPBW"/>
       <xs:enumeration value="SwitchModal"/>
+      <xs:enumeration value="PasswordOnly"/>
```
rosswbrown, line 53: Update `AuthOptionEnum-ST` below as well with `Password`

saugatt: Made the change in the file.

PR #557, opened 2020-07-13
Update tdvt for handling password file for server version test

jainam1995:
Update the tdvt parsing logic to handle password file in server version tests

### DIFF
--- a/tdvt/tdvt/tabquery.py
+++ b/tdvt/tdvt/tabquery.py
@@ -41,13 +41,13 @@ def build_tabquery_command_line(work):
     cmdline = tb.build_tabquery_command_line(work)
     return cmdline
 
-def build_connectors_test_tabquery_command_line(connTestName, connTestFileName, connTestPasswordFile):
+def build_connectors_test_tabquery_command_line(conn_test_name, conn_test_file_name, conn_test_password_file):
     global tab_cli_exe
     cmdline = [tab_cli_exe]
-    cmdline.extend(["--conn-test", connTestName])
-    cmdline.extend(["--conn-test-file", connTestFileName])
-    if connTestPasswordFile:
-        cmdline.extend(["--conn-test-password-file", connTestPasswordFile])
+    cmdline.extend(["--conn-test", conn_test_name])
+    cmdline.extend(["--conn-test-file", conn_test_file_name])
+    if conn_test_password_file:
+        cmdline.extend(["--conn-test-password-file", conn_test_password_file])
     return cmdline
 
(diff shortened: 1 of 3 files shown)


tdvt/tdvt/tabquery.py
```diff
@@ -41,11 +41,13 @@ def build_tabquery_command_line(work):
     cmdline = tb.build_tabquery_command_line(work)
     return cmdline
 
-def build_connectors_test_tabquery_command_line(connTestName, connTestFileName):
+def build_connectors_test_tabquery_command_line(connTestName, connTestFileName, connTestPasswordFile):
```
lukewrites, line 44:
```suggestion
def build_connectors_test_tabquery_command_line(connTestName, connTestFileName, connTestPasswordFile=None):
```
I think we need this to default to None otherwise it'll blow up without a password file.

tdvt/tdvt/tdvt.py
```diff
@@ -417,7 +417,9 @@ def get_level_of_parallelization(args):
     '''
 run_connectors_test_usage_text = '''
     The commands below can be used to run the connectors tests.
-    'filename.xml' in the below commands is the path of the xml test file used to run the tests.
+    'filename.txt' in the below commands is the path of the xml test file used to run the tests.
```
lukewrites, line 420: Was there a reason this went back to `filename.txt` rather than `.xml`?

tdvt/tdvt/tdvt_core.py
```diff
@@ -32,13 +32,14 @@
 
 
 class ConnectorsTest(object):
-    def __init__(self, conn_test_name, conn_test_file):
+    def __init__(self, conn_test_name, conn_test_file, conn_test_password_file):
```
lukewrites, line 35: Can you verify this will work without a password file specified? Similar comment to above, probably need 
```suggestion
    def __init__(self, conn_test_name, conn_test_file, conn_test_password_file=None):
```

jainam1995: Yes i had tested this case before sending out the PR. It works correctly with and without the password file

tdvt/tdvt/tdvt.py
jainam1995: Oops, git mis match. updating it

tdvt/tdvt/tabquery.py
jainam1995: I tested with and without the password file. The current code works correctly

jainam1995:
> Couple of small things and a nit 😄

Fixed

PR #565, opened 2020-07-28
Changing the Staples smoke tests to be logical to allow column name c…

icooptableau:
…hanges

tfs1021546

**Things To Remember:** 

* **`dev-2020.1`**
* **`2020.2`**
=====


lukewrites:
LGTM, please update `tdvt/CHANGELOG.md` with this change in the `[Unreleased]` section.

icooptableau:
Updated the changelog

tdvt/tdvt/exprtests/pretest/connection_tests/staples/setup.staples_connection_test.xml
```diff
@@ -0,0 +1,220 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<logical-query>
+  <selects>
+    <field>[PID]</field>
+    <field>[sum:Gross Profit:ok]</field>
+    <field>[sum:Sales Total:ok]</field>
+  </selects>
+  <domainOp class="logical-operator" type="snowflake">
```
rosswbrown, line 8: So this is saying a good smoke test for the Staples table being loaded correctly is roughly (forgive my rusty SQL):

SELECT 'PID', SUM('Gross Profit'), SUM('Sales Total')
FROM 'Staples'
WHERE 'Call Center Region' == EAST && 'Order Date' == 2002
GROUP BY 'PID'
HAVING ('Sales Total' > -3640.23 && 'Sales Total' < 24622.400000000001) 

Did we pull this from somewhere else or an evaluation you did of what data from Staples is best to check?

Before this change what was checked?

icooptableau: Honestly I just copied a similar Staples logical test to use as a smoke test.  It was just to validate that I could get a logical smoke test to run.  If there is a better test I'm all for it but it looks like it's rarely used in the connector ini files.  

What I checked before hand was that I could successfully run the smoke tests against postgres before the change, then after made sure it could load the logical tests in that format.

tdvt/tdvt/config_gen/datasource_list.py
```diff
@@ -327,7 +327,7 @@ def load_test(config, test_dir=get_root_dir()):
 
             try:
                 all_ini_sections.remove(section)
-                test_config.add_expression_test('StaplesConnectionTest', STAPLES_TDS, sect.get(KEY_EXCLUSIONS, ''),
+                test_config.add_logical_test('StaplesConnectionTest', STAPLES_TDS, sect.get(KEY_EXCLUSIONS, ''),
```
rosswbrown, line 330: So this is needed to account for the different LogicalQueryFormat configurations?

rosswbrown: The answer is apparently yes, please note in the item which config is used and the partner details.

PR #570, opened 2020-08-11
Add priority and category metadata files

erademacher:
Added two new CSV files.

test_metadata.csv contains a denormalized mapping of tests to categories and priorities. In a future change, it will be read by the TDVT output formatter and joined with the test results.

priority.csv is a normalized map of tests to priorities, which is included because it's more compact and easily analyzable by plugin developers in the interim before this info is included in TDVT output.

None of the prioritization is scientific in any way. These are just the priorities I would assign to tests if I were analyzing a vendor's results. Discussion and debate is welcome.


tsjoblad:
Hey Evan, this is great! Very excited about how we will improve our partner's development experience, so thank you for the work here. I'd love to be included in any conversations the team has about how these mappings will change or be honed. I have a few quick thoughts:
- I see a number of date tests listed as recommended. My impression is that since we need values to stay constant between extracts and live connections, shouldn't these be required?
- Right now the percentage of required tests is at 65%, or 76% if you include the relational tests. This is well below our team's benchmark of 80% needing to pass before a review. This seems a bit dissonant and I'd love to dig into this a bit more. Thanks!

tdvt/tdvt/metadata/priority.csv
```diff
@@ -0,0 +1,327 @@
...
+Filter.slicing_Q_date_month_instance_filter	Required
+Filter.TopN-context	Required
+Filter.Trademark	Smoke
+join.null.bool	Recommended
+join.null.date	Recommended
+join.null.datetime	Recommended
+join.null.int	Recommended
+join.null.real	Recommended
+join.null.str	Recommended
+lod.1_Difference From with Cross Join	Required for Relational
```
pvanderknyff, line 198: I think at the moment we only support relational databases, there could be an argument for making these fully required to avoid confusion (if we only ever intend to support relational and not other types of database)

tdvt/tdvt/metadata/priority.csv
```diff
@@ -0,0 +1,327 @@
...
+date.dateadd.weekday	Required
+date.dateadd.year	Required
+date.datediff.day	Required
+date.datediff.dayofyear	Required
+date.datediff.hour	Required
+date.datediff.iso-week	Recommended
+date.datediff.iso-weekday	Recommended
+date.datediff.minute	Required
+date.datediff.month	Required
+date.datediff.nulls	Recommended
```
rosswbrown, line 60: I think assuming the data source supports nulls with date/datetime then I think we need something like "Required if supported"

pvanderknyff:
Outside of this story, but updating our [tdvt results workbook](https://github.com/tableau/connector-plugin-sdk/blob/master/tdvt/TDVT%20Results.twbx) to use this information would be great.

tsjoblad:
> Outside of this story, but updating our [tdvt results workbook](https://github.com/tableau/connector-plugin-sdk/blob/master/tdvt/TDVT%20Results.twbx) to use this information would be great.

I can do that once this is merged/finalized!

tdvt/tdvt/metadata/priority.csv
erademacher: Maybe I'll change "Required for relational" to a more generic "Required if supported".

tdvt/tdvt/metadata/priority.csv
erademacher: By "relational", I mean actual ACID-compliant relational databases, not hadoop or NoSQL databases hiding behind a SQL engine. But I should change this to "Required if supported" if it makes that a little more clear.

PR #573, opened 2020-08-12
Clarify connection-properties description

pvanderknyff:
Jade pointed out that this was confusing and made it seem like the bug was still active, and considering how long it took me to parse what this entry was trying to say despite knowing the topic I agree.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -42,7 +42,7 @@ Similar to connection-builder but is used to build the JDBC properties file. For
 
 **Type:** JavaScript
 
-#### JavaScript function call signature (Recomended, added in 2019.4.1):
+#### JavaScript function call signature (Recommended, added in 2019.4.1):
 
 **Input:** attr, an object of key/value pairs
 
@@ -68,7 +68,7 @@ Similar to connection-builder but is used to build the JDBC properties file. For
 **Return:** an array of formatted key=value pairs that will be written to the properties file
 
 ```javascript
-{"UID=myusername", "Host=myserver.somewhere.net", "PWD=mypassword"};
+["UID=myusername", "Host=myserver.somewhere.net", "PWD=mypassword"]
 ```
 
 ---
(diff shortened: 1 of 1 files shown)


docs/api-reference.md
```diff
@@ -39,30 +39,36 @@ Builds the ODBC ConnectString or JDBC Connection URL. For a JDBC Connection URL,
...
-
-an object of key/value pairs that will be written to the properties file(recommended and available since 2019.4.1)
+**Return:** an object of key/value pairs that will be written to the properties file(recommended and available since 2019.4.1)
 
 ```javascript
 ["UID" : "myusername", "Host" : "myserver.somewhere.net", "PWD" : "mypassword"];
 ```
 
-array of formatted key=value pairs that will be written to the properties file
+#### JavaScript function call signature (Legacy):
```
rosswbrown, line 59:
```suggestion
#### JavaScript function call signature (Deprecated):
```
It means is still there, not recommended, and can be removed at any time.

docs/api-reference.md
```diff
@@ -39,30 +39,36 @@ Builds the ODBC ConnectString or JDBC Connection URL. For a JDBC Connection URL,
 ### connection-properties
 
 Similar to connection-builder but is used to build the JDBC properties file. For JDBC Connection URL, we require that connection-properties contain secure attributes such as username and password.
-**Note:** A bug in the JavaScript translation layer means that you cannot use values containing the '=' character.
 
 **Type:** JavaScript
 
-#### JavaScript function call signature:
+#### JavaScript function call signature (Recomended, added in 2019.4.1):
```
jkoskela, line 45: Recommended

docs/api-reference.md
```diff
@@ -39,30 +39,36 @@ Builds the ODBC ConnectString or JDBC Connection URL. For a JDBC Connection URL,
...
+
+```javascript
+{"server" : "myserver.somewhere.net", "username" : "myusername", "password" : "mypassword"}
+```
+
+**Return:** an array of formatted key=value pairs that will be written to the properties file
 
 ```javascript
-{"UID" : "myusername", "Host" : "myserver.somewhere.net", "PWD" : "mypassword"};
+{"UID=myusername", "Host=myserver.somewhere.net", "PWD=mypassword"};
```
jkoskela, line 71: Should use square brackets.

PR #581, opened 2020-08-14
Add mcd documentation to index

pvanderknyff:
I'm not sure the best way to differentiate between the old style of connection dialogs and modular dialogs, which we call Connection Dialogs v2 externally. Calling the old syle "Legacy", should it be "deprecated"?

Diff looks bigger than it is since my VSCode removes trailing spaces on save.

### DIFF
--- a/_includes/docs_menu.html
+++ b/_includes/docs_menu.html
@@ -7,7 +7,7 @@
     <li>Basic Concepts</li>
     <li><a href="{{ site.baseurl }}/docs/design">Design Considerations</a></li>
     <li><a href="{{ site.baseurl }}/docs/dialect">Create a Dialect Definition (TDD)</a></li>
-    <li><a href="{{ site.baseurl }}/docs/ui">Build the Connection Dialog</a></li>
+    <li><a href="{{ site.baseurl }}/docs/ui">Build the Connection Dialog with Connection Dialog v1</a></li>
     <li><a href="{{ site.baseurl }}/docs/mcd">Build the Connection Dialog with Connection Dialog v2</a></li>
     <li><a href="{{ site.baseurl }}/docs/tdvt">Test with the TDVT Suite</a></li>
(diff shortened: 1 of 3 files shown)


docs/mcd.md
```diff
@@ -1,7 +1,7 @@
 ---
-title: Connection Dialog v2
---- 
-**IMPORTANT:** This feature is only available in Tableau 2020.3 or later.
+title: Build the Connection Dialog with Connection Dialog v2
+---
+**IMPORTANT:** This feature is only available in Tableau 2020.3 or later. For compatibility with older versions, use the legacy connection dialogs (documentation found [here]({ site.baseurl }}/docs/ui)).
```
sxlin, line 4: Call it Connection Dialog V1?

rosswbrown: +1, its still supported so not legacy.

dmoynihan: Please change "This feature is only available in" to "This feature is available only in". (Only modifies set of versions, not "available".) Thanks.

pvanderknyff: Updated, thank you! Sorry for the long delay, this pull request completely slipped my mind.

_includes/docs_menu.html
```diff
@@ -7,7 +7,8 @@
     <li>Basic Concepts</li>
     <li><a href="{{ site.baseurl }}/docs/design">Design Considerations</a></li>
     <li><a href="{{ site.baseurl }}/docs/dialect">Create a Dialect Definition (TDD)</a></li>
-    <li><a href="{{ site.baseurl }}/docs/ui">Build the Connection Dialog</a></li>
+    <li><a href="{{ site.baseurl }}/docs/ui">Build the Connection Dialog with Connection Dialog v1</a></li>
```
rosswbrown, line 10: If the sidebar starts to look weird with long titles we could try "Build the Connection Dialog with 1.0"/"... with 2.0".  We can always iterate.

PR #589, opened 2020-08-25
Update initialsql tests to use RAWSQLAGG_STR instead of RAWSQL_STR

IanTamTableau:
The update is needed because a fix involving RAWSQL_STR was causing the initialsql tests to fail on StandardDB.


IanTamTableau:
@atalukdar123

IanTamTableau:
@atalukdar123 I have tested my recent changes with the initialsql tests on the following databases, and confirmed that they pass on my local machine : vertica, memsql, db2, googlecloudsql, hana, mariadb, mysql, netezza, paraccel, postgres, redshift_multinode, snowflake, sqlserver, sybase_ase, vertica.

PR #610, opened 2020-09-17
Add all_options.ini sample

pvanderknyff:
The documentation for TDVT has a gigantic 90-line ini file as the only documentation for how to make one of these. It's large and intimidating, and is hard to parse when trying to figure out something specific.

As part of my TDVT doc update, I'm rewriting that section and removing this giant blob of text. However, this may still be useful to people to see the syntax or something, so I'm adding it as a standalone example file.

### DIFF
--- a/tdvt/samples/config/all_options_example.ini
+++ b/tdvt/samples/config/all_options_example.ini
@@ -1,0 +1,90 @@
+[Datasource]
+
+Name = your_datasource_name  #i.e. mydb
+
+CommandLineOverride = -DLogLevel=Debug -DConnectPluginsPath=[path_to_unpackaged_connector] #Space separated list of arguments that are passed through unchanged to tabquerytool. Most Tableau arguments require a prepended '-D'.
+#If you are testing a connector, be sure to add the command line argument -DConnectPluginsPath and have it pointed at the folder where your connector is located
+
+MaxThread = 6   #You can add this to control Max Thread number when you use TDVT to run single datasource, it cannot apply with multi datasource
+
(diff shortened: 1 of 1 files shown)


rosswbrown (approved):

tdvt/samples/config/all_options_example.ini
```diff
@@ -0,0 +1,90 @@
...
+# They can be run by themselves using the --verify flag (e.g. tdvt run postgres --verify).
+CastCalcsTestEnabled = True
+StaplesTestEnabled = True
+
+[LODTests]
+
+[UnionTest]
+
+#Advanced:
+#You can add a new arbitrary test suite like this.
```
tsjoblad, line 72: Maybe add something like, this will likely not be needed for regular TDVT usage and thus everything below line 70 is not recommended?

tdvt/samples/config/all_options_example.ini
```diff
@@ -0,0 +1,90 @@
...
+Name = your_datasource_name  #i.e. mydb
+
+CommandLineOverride = -DLogLevel=Debug -DConnectPluginsPath=[path_to_unpackaged_connector] #Space separated list of arguments that are passed through unchanged to tabquerytool. Most Tableau arguments require a prepended '-D'.
+#If you are testing a connector, be sure to add the command line argument -DConnectPluginsPath and have it pointed at the folder where your connector is located
+
+MaxThread = 6   #You can add this to control Max Thread number when you use TDVT to run single datasource, it cannot apply with multi datasource
+
+TimeoutSeconds = 6000 # You can overload the timeout for queries if your queries are running slow. By default this is set to 3600.
+
+LogicalQueryFormat = dbo
```
tsjoblad, line 12: my_logical_query?

PR #616, opened 2020-09-22
Sort test results by complexity (number of functions tested)

erademacher:
Added an intermediary list to store test results instead of copying straight to the final output. Lists in python are thread-safe, so there shouldn't be any threading issues. The results are then sorted by the distinct number of functions tested.


rosswbrown (approved):
Looks good.  Can you ensure this version makes it into near as well.  Its currently using 2.1.18. That is the only other validation I can think of: no regression in the TeamCity views that consume this.

PR #621, opened 2020-09-28
Fix metadata file path

erademacher:
Fixed path resolution for the metadata file and added error handling in case it's missing


lukewrites:
Could you run the unit tests to see if they pass unchanged after this? If they do we probably need to add a test involving a metadata file.

pvanderknyff (approved):
LGTM, though I agree with Luke that we should probably have a unit test covering this so if it gets broken again the pipeline will catch it.

erademacher:
Sorry, I just don't have time to write a test this sprint. I'm going to push it through (with a version bump) and put something in TFS for test debt.

PR #623, opened 2020-09-30
JDBC Kerberos Postgres sample using Modular Connection Dialog

devanshsoni9:
Tested manually on Windows Desktop + Server, Mac Desktop, Linux Server using instructions documented in the Readme

Corresponds to TFS1104877


samples/scenarios/jdbc_kerberos/README.md
```diff
@@ -0,0 +1,152 @@
+
+# Readme
+
+The postgres jdbc kerberos sample connector plugin demonstrates how to incorporate different kerberos-based authentication modes for your "Tableau + Postgres + JDBC" use cases. This sample plugin code has been tested for all the authentication methods listed here with all platforms (Mac Desktop, Windows Desktop and Server, Linux Server). However, the individual sections of the plugin components concern specific authentication modes and have been laid out here for the purposes of understanding and re-use. While publishing from Tableau Desktop, the Server RunAs and Viewer credentials authentication options will be visible in publish dialog only if you have logged into tableau using Integrated Authentication.
```
jkoskela, line 4: "While publishing from Tableau Desktop, the Server RunAs and Viewer credentials authentication options will be visible in publish dialog only if you have logged into tableau using Integrated Authentication." 

This part is no longer true. You should get that option as long as you enable wgserver.delegation.enabled and the CAP_KERBEROS_IMPERSONATE is enabled.

samples/scenarios/jdbc_kerberos/README.md
```diff
@@ -0,0 +1,152 @@
...
+1. Use **SSPI** authentication to connect a postgres database with `Tableau Desktop on Windows`   
+1. Use **Kerberos SSO via GSS-API** to connect a postgres database with `Tableau Desktop on Mac`.
+1. Publish a viz from `Tableau Desktop` using **Server RunAs authentication** and open it in Tableau Server. In this scenario, Tableau Server connects to databases with a service account.
+1. Publish a viz from `Tableau Desktop` using **Kerberos Delegation/Viewer Credentials authentication** and open it in Tableau Server. When a user accesses this viz on Tableau Server, the impersonation_runas_principal would be used to delegate queries to the tableau server user. 
+
+Below I give a brief introduction on how Java uses GSS-API to implement Kerberos authentication.
+
+## <a id="intro"> Introduction to Kerberos on Java
+
+### GSS-API
```
jkoskela, line 23: I'm not sure if explaining the difference between sspi and gssapi is necessary now. It might just confuse them since Postgres JDBC has an option for sspi, but they should always use gssapi. 

You might just add a note that says to always use gssapi, and that on Windows Tableau uses sspi_bridge to translate from gssapi to sspi. And then they don't have to spend much time thinking about it.

samples/scenarios/jdbc_kerberos/README.md
```diff
@@ -0,0 +1,152 @@
...
+-   for authorization of users to ensure they have the access control rights (permissions) required to do security-sensitive operations.
+
+Java GSS-API is used for securely exchanging messages between communicating applications. The Java GSS-API contains the Java bindings for the GSS-API defined in [RFC 5653](https://tools.ietf.org/html/rfc5653). 
+(source - [Oracle Java docs](https://docs.oracle.com/en/java/javase/11/security/java-generic-security-services-java-gss-api1.html)).  
+
+JAAS authentication is typically performed prior to secure communication using Java GSS-API. Tableau Server uses JAAS and JGSS for Server Runas and Kerberos Delegation authentication.
+
+## <a id="desktop"></a>Kerberos SSO for Tableau Desktop
+
+### SSPI authentication for Tableau Desktop on Windows
```
jkoskela, line 47: This is now the same as Mac, no reason for separate sections.

samples/scenarios/jdbc_kerberos/README.md
```diff
@@ -0,0 +1,152 @@
...
+    props["jaasApplicationName"] = "com.sun.security.jgss.krb5.initiate";
+    props["gsslib"] = "gssapi";	 
+    props["jaasLogin"] = "false";  
+```
+
+Other Configuration:
+* Only for postgres jdbc plugin on Mac - add this property `Settings.DisableNativeGSS` as a `Boolean/YES` to the Tableau plist on Mac at `/Library/Preferences/com.tableau.Tableau-<version>.plist` as described below. Here version is Tableau Desktop major version, for example, `com.tableau.Tableau-2020.4.plist`.
+```
+// Note - the below command requires root privileges hence sudo is needed.
+sudo defaults write /Library/Preferences/com.tableau.Tableau-<version>.plist Settings.DisableNativeGSS -bool YES
```
jkoskela, line 88: Did you ever open a bug on Postgres JDBC for that?

samples/scenarios/jdbc_kerberos/README.md
```diff
@@ -0,0 +1,152 @@
...
+```
+* Make sure a krb5.conf file is present at `/etc/krb5.conf`(/etc is a private directory, requires root privileges). Check for the existence of either of the following two files `/etc/krb5.conf` or `/Library/Preferences/edu.mit.Kerberos`. If the second file (edu.mit.Kerberos) is present it needs to be backed up and deleted.
+* Before connecting, run `klist` to verify kerberos TGT for the user principal is present in the kerberos ticket cache. If not, run `kinit user@REALM` before opening Tableau Desktop.  
+
+
+## <a id="server"/> Kerberos RunAs and Delegation on Server
+When a user logs into Tableau desktop using Integrated authentication, there are the two publish options available for Kerberos authentication, namely, "Server run-as" and "Viewer Credentials/Kerberos Delegation".
```
jkoskela, line 95: Logging in using integrated auth no longer required. Where did you see that it was required?

samples/scenarios/jdbc_kerberos/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+connectionProperties.js
+```
+    function isEmpty(str) {
+        return (!str || 0 === str.length); 
+    }
+
+    if (attr[connectionHelper.attributeAuthentication] == "auth-integrated") {
+        // if attributeTableauServerUser is non-empty, it means the connector plugin is currently being accessed in a tableau server environment
+        var serverUser = attr[connectionHelper.attributeTableauServerUser];
```
jkoskela, line 113: This is the run as user here right? Might want to add a comment to explain that. Also this example is the same as below, maybe just use one code example with the comment that explains that in the run-as case the tableauServerUser which is passed down is the runAsUser.

samples/scenarios/jdbc_kerberos/README.md
jkoskela, line 4: Not a big deal, but next time try to use consistent capitalization for public facing stuff like "Postgres JDBC"

jkoskela (approved):

samples/scenarios/jdbc_kerberos/README.md
devanshsoni9: I did not open a bug on them. I should probably do that. This is not a great experience.

samples/scenarios/jdbc_kerberos/README.md
devanshsoni9: I did not know we fixed that. I will test without that to check it works fine. Thanks

samples/scenarios/jdbc_kerberos/README.md
devanshsoni9: yes, I almost forgot about that. This is the [issue](https://github.com/pgjdbc/pgjdbc/issues/1662) which I reported some time back with 42.2.5 driver. Now we are using 42.2.14 and it is still not been fixed.

samples/scenarios/jdbc_kerberos/README.md
devanshsoni9: Tested this and changed the wording so that user is not confused.
New wording

"When a user opens Tableau Desktop and login to Postgres using Kerberos there are the two publish options available"

samples/scenarios/jdbc_kerberos/README.md
devanshsoni9: removed this section entirely. 
It looks cleaner now 
https://github.com/tableau/connector-plugin-sdk/tree/dsoni_postgres_kerb/samples/scenarios/jdbc_kerberos

PR #631, opened 2020-10-13
Update documentation regarding connectionMatcher and connectionNormal…

jainam1995:
Update documentation as we have deprecate connectionMatcher.js and connectionNormalizer.js files.


jainam1995:
files-overview
files-sequence

docs/api-reference.md
```diff
@@ -229,6 +211,9 @@ Example:
 
 ## Deprecated API
 
+### connectionNormalizer.js and connectionMatcher.js
+These .js files are deprecated as of Tableau 2020.3. The <connection-normalizer> component can be added to the connectionResolver.tdr fie as shown in the connection-normalizer section above.
```
rosswbrown, line 215: Can you add the element/attribute names.  Technically the files could have been named anything before

docs/example.md
```diff
@@ -68,10 +68,28 @@ The Connector Resolver file (.tdr) is optional. Tableau uses it to create a conn
...
+            <attr>port</attr>
+            <attr>dbname</attr>
+            <attr>username</attr>
+            <attr>password</attr>
+        </attribute-list>               
+    </required-attributes>
+</connection-normalizer>
+
+```
+`<setImpersonateAttributes/>` adds support for impersonate attributes and should be in every connector.
```
rosswbrown, line 89: `setImpersonateAttributes` can be removed, its done automatically.  I believe samples are already updated.

docs/example.md
```diff
@@ -82,7 +100,40 @@ If the attributes and their values are identical then the connections are consid
...
-![]({{ site.baseurl }}/assets/pce-tdr.png)
+```
+<tdr class='postgres_odbc'>
+    <connection-resolver>
+        <connection-builder>
+            // This is the component number 4 in the diagram above
+            <script file='connectionBuilder.js'/> 
+        </connection-builder>
+        <connection-properties>
+            // This is the component number 5 in the diagram above
```
rosswbrown, line 111: Can you add JDBC only in the comment here too

rosswbrown (commented):
We should remove the arrow and { related to the dialect in the images.  Visually it implies the dialect is used to connect.  There are narrow cases when this is true at an implementation level, however the dominate case is for query formation, which occurs after initial connection.  The description in 7 below I think articulates this well.

docs/api-reference.md
jainam1995: Added

docs/example.md
jainam1995: Removed

docs/example.md
jainam1995: Added

jainam1995:
> We should remove the arrow and { related to the dialect in the images. Visually it implies the dialect is used to connect. There are narrow cases when this is true at an implementation level, however the dominate case is for query formation, which occurs after initial connection. The description in 7 below I think articulates this well.

Updated the images. Adding screenshot of the new images in comments below

jainam1995:
files-overview
files-sequence

jainam1995:
Addressed @rosswbrown comments

docs/example.md
```diff
@@ -185,7 +245,7 @@ For JDBC, search for <span style= "font-family: courier new">Connection URL</spa
 JDBCProtocol Connection URL: jdbc:postgresql://postgres:5342/TestV1?user=test&password=********
```
rosswbrown, line 245:
```suggestion
JDBCProtocol Connection URL: jdbc:postgresql://postgres:5342/TestV1
```
This example pre-dated the propertiesBuilder.  user/pass are no longer passed on the url.

rosswbrown (approved):

docs/example.md
jainam1995: Fixed

PR #641, opened 2020-10-28
New TDVT Tests to deprecate Python tests and add some ADD, DIFF and AGG test cases.

krhodes-tableau:
(First time checking in; please Slack me if you have any suggestions and/or additional review requirements to encourage the success of TDVT additions)

Part of the Python deprecation effort is to remove the test cases that are currently running in Python 2.7 and push them to other more appropriate frameworks. The following test cases being added serve two purposes; they remove the slow test cases that run in the full product and use the much faster TDVT framework, and they provide better (slightly unnecessary) coverage by being run against multiple data sources.  The real benefit here apart from deprecation is the speed improvement.

Some minor updates to the existing files were also made to address a bug that was added (but exercised the same code, so never updated) in the initial merge. I suspected long ago the person that would have to address that would be me and this became a self fulfilling prophecy. 

Relevant command used to validate the results:
`python tdvt_wrapper.py run-pattern sqlserver --exp exprtests/iso8601week/* --tdp cast_calcs.*.tds`

Result:

> Test Count: 189 tests
>         Passed tests: 189
>         Failed tests: 0
>         Tests run: 189
>         Disabled tests: 0
>         Skipped tests: 0

Special thanks to the following for helping achieve this goal: @lriggs, @rosswbrown, @erademacher, @lukewrites. All ISO8601 data sources should support these tests.


lukewrites:
@krhodes-tableau the tests look good, I'll test them on local, but you'll also want to bump the version number of tdvt so this can be packaged and consumed by TC.

lukewrites:
image
LGTM

lukewrites (approved):
Changes look good; just need a version bump in order to merge.

krhodes-tableau:
My git-fu had trouble with the commit-push dance.  Version updated to 2.2.1 in both the changelog and the version.py files. Please especially scrutinize those two files closely before merging to master.

lukewrites (approved):
:shipit: 
looks great, thanks!

PR #655, opened 2020-12-02
[SDK] TDD definition needs to say that format is distinct can give the wrong answer

jainam1995:
**Things To Remember:** 

* **`dev-2020.1`**
* **`2020.2`**
=====

### DIFF
--- a/docs/dialect.md
+++ b/docs/dialect.md
@@ -159,7 +159,7 @@ Allowable date parts: year, quarter, month, dayofyear, day, weekday, week, hou
  The value for this property can be:
 
   - __NoNullCheck__
-      Does not check that the value of LHS and RHS is null. Logic: `(lhs [!]= rhs)`
+      Does not check that the value of LHS and RHS is null. Logic: `(lhs [!]= rhs)` . Note : A known limitation is that using this value will cause Tableau to give the wrong answer in some cases, if the column is ever null. It is adviced not to use this value unless there is no other option.
   - __Keyword__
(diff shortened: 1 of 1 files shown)


dcory (commented):
Don't you mean "advised" instead of "adviced"?

jainam1995:
> Don't you mean "advised" instead of "adviced"?

Yes, updated to advised.

docs/dialect.md
```diff
@@ -159,7 +159,7 @@ Allowable date parts: year, quarter, month, dayofyear, day, weekday, week, hou
  The value for this property can be:
 
   - __NoNullCheck__
-      Does not check that the value of LHS and RHS is null. Logic: `(lhs [!]= rhs)`
+      Does not check that the value of LHS and RHS is null. Logic: `(lhs [!]= rhs)` . Note : A known limitation is that using this value will cause Tableau to give the wrong answer in some cases, if the column is ever null. It is advised not to use this value unless there is no other option.
```
rosswbrown, line 162:
```suggestion
      Does not check that the value of LHS and RHS is null. Logic: `(lhs [!]= rhs)` . Note: A known limitation is that using this value will cause Tableau to give the wrong answer in some scenarios if the column contains null. This is not recommended unless required by the underlying database.
```

Someone else can please double check my wording too, thanks

Another way to make this very clear, make it the last in the list instead of first, and change the example on line 177 to `Operator`.

jainam1995: Fixed

rosswbrown (approved):

Issue #640, opened 2020-10-28
Quote column names in DDL scripts


redcatbear:
**Is your feature request related to a problem? Please describe.**
Column names in DDL scrips can conflict with reserved names depending on the database the DDL is us executed on. The word "YEAR" for example is a reserved word in Exasol.

**Describe the solution you'd like**
Wrapping the column names in the DDL script in quotes would keep the column names intact while at the same time clearly indicating that the are identifiers. This resolve any potential keyword conflict.

**Describe alternatives you've considered**
Renaming columns would break existing tests and there is no guarantee that a renamed column would not conflict with a different database.

**About You**
Name: @redcatbear 
Company: Exasol AG

redcatbear:
Examples for reserved words in https://github.com/tableau/connector-plugin-sdk/tree/master/tests/datasets/TestV1
- `year`
- `cs`
- `key`

sidwray:
tfs 1197768

jsun28:
Column name quoting was done in this PR: https://github.com/tableau/connector-plugin-sdk/pull/674, so I am going to close this.
also, our DDL scripts were meant to be an example for individuals to modify as needed to meet their needs, so if you have other needs in the future, you can also modify them yourself.

PR #674, opened 2020-12-17
Add quotes to DDL scripts' column names

jsun28:
This change is for [TFS 1197768](https://tfs.tsi.lan/tfs/DefaultCollection/Tableau-Dev/_workitems/edit/1197768) and [github issue 640](https://github.com/tableau/connector-plugin-sdk/issues/640).

The customer is complaining about the column names in our DDL scrips conflict with their reserved names.

Quoting those column names can indicate that those are identifiers then resolve the conflict, also it won't change the column names, I talked to Logan and manually tested it myself, it won't affect anything since column names are the same as before.


rosswbrown (approved):
Yeah, this is meant to be an example for individuals to modify as needed to meet their needs.  Quotes are more clear and match Staples as well.

PR #687, opened 2021-01-07
Support packaging OAuth connectors [wip]

pvanderknyff:
Merging in dev-2021.1 branch right now, but can merge into a separate dev-oauth-wip branch if people think that's a better idea.

### DIFF
--- a/connector-packager/connector_packager/jar_jdk_packager.py
+++ b/connector-packager/connector_packager/jar_jdk_packager.py
@@ -41,10 +41,12 @@ def get_min_support_version(file_list: List[ConnectorFile]) -> str:
     for connector_file in file_list:
         # if we have a connection-fields file, then we are using modular dialogs and need 2020.3+
         if connector_file.file_type == "connection-fields":
-            min_version_tableau = "2020.3"
+            if 2020.3 > float(min_version_tableau):
+                min_version_tableau = "2020.3"
             reasons.append("Connector uses Connection Dialogs V2, which was added in the 2020.3 release")
         elif connector_file.file_type == "oauth-config":
-            min_version_tableau = "2021.1"
+            if 2021.1 > float(min_version_tableau):
+                min_version_tableau = "2021.1"
             reasons.append("Connector uses OAuth, which was added in the 2021.1 release")
 
     return min_version_tableau, reasons
--- a/connector-packager/connector_packager/jar_jdk_packager.py
+++ b/connector-packager/connector_packager/jar_jdk_packager.py
@@ -43,6 +43,9 @@ def get_min_support_version(file_list: List[ConnectorFile]) -> str:
         if connector_file.file_type == "connection-fields":
             min_version_tableau = "2020.3"
             reasons.append("Connector uses Connection Dialogs V2, which was added in the 2020.3 release")
+        elif connector_file.file_type == "oauth-config":
+            min_version_tableau = "2021.1"
+            reasons.append("Connector uses OAuth, which was added in the 2021.1 release")
 
     return min_version_tableau, reasons
 
--- a/connector-packager/tests/test_resources/oauth_connector/connectionFields.xml
+++ b/connector-packager/tests/test_resources/oauth_connector/connectionFields.xml
@@ -7,18 +7,24 @@
 
   <field name="port" label="Port" value-type="string" category="endpoint" default-value="5432" />
 
-  <field name="v-custom" label="Custom" value-type="string" category="endpoint" />
+  <field name="authentication" label="Auth" category="authentication" value-type="selection" default-value="oauth">
+        <selection-group>
+            <option label="OAuth" value="oauth"/>
+            <option label="Username/Password" value="auth-user-pass" />
+        </selection-group>
+   </field>
 
-  <field name="username" label="Username" value-type="string" category="authentication" />
+   <field name="username" label="Username" category="authentication" value-type="string">
+        <conditions>
+            <condition field="authentication" value="auth-user-pass" />
+        </conditions>
+    </field>
 
-  <field name="password" label="Password" value-type="string" category="authentication" secure="true" />
-  
-  <field name="v-custom2" label="Advanced Custom" value-type="string" category="advanced" optional="false" default-value="test" />
-  
-  <field name="vendor1" label="Advanced Vendor1" value-type="string" category="advanced" optional="false" default-value="testVendor1" />
-  
-  <field name="vendor2" label="Advanced Vendor2" value-type="string" category="advanced" optional="false" default-value="testVendor2" />
-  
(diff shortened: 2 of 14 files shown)


validation/oauth_config.xsd
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rosswbrown, line 1: We'll want a test case that uses this xsd/file test_xml_parser.py and/or test_xsd_validator.py I think.

connector-packager/connector_packager/jar_jdk_packager.py
```diff
@@ -43,6 +43,9 @@ def get_min_support_version(file_list: List[ConnectorFile]) -> str:
         if connector_file.file_type == "connection-fields":
             min_version_tableau = "2020.3"
             reasons.append("Connector uses Connection Dialogs V2, which was added in the 2020.3 release")
+        elif connector_file.file_type == "oauth-config":
```
rosswbrown, line 46: You'll want the newest check first so we end up computing the max min_version_tableau.

validation/oauth_config.xsd
rosswbrown: I was looking at an out of date tab, you can possibly ignore this.

pvanderknyff: Added a test case after initial version that tests this, using the existing mcd test as a guide.

connector-packager/connector_packager/jar_jdk_packager.py
pvanderknyff: This does end up correctly stamping 2021.1 (the test checks for this), but the reason it does it the connection fields file is checked first (since it's first in the manifest), so it's kind of by accident. I think only changing the min_version if it's greater than the current one is probably safest.

connector-packager/tests/test_resources/oauth_connector/oauth-config.xml
```diff
@@ -0,0 +1,34 @@
...
+
+    <authUri>/oauth2/v2.0/authorize</authUri>
+    <tokenUri>/oauth2/v2.0/token</tokenUri>
+
+    <scopes>openid</scopes>
+    <scopes>email</scopes>
+    <scopes>profile</scopes>
+    <scopes>offline_access</scopes>
+
+    <!-- TODO: required attrs need review -->
```
wangxinlei, line 21: Line 21-33 can be removed

validation/oauth_config.xsd
wangxinlei, line 1: How would we keep this consistent with https://gitlab.tableausoftware.com/connectivity/oauth-service/-/blob/master/src/main/resources/xsd/oauthConfig.xsd? Is it just copy the changes by hand currently?

rosswbrown: Yes, within the dev-<release> branch.  We also should audit on release boundaries like we do with capabilities today as well in case a change gets missed.

pvanderknyff: Yes, unfortunately.

connector-packager/tests/test_resources/oauth_connector/oauth-config.xml
pvanderknyff: Removed.

connector-packager/tests/test_resources/oauth_connector/connectionFields.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
```
rosswbrown, line 1: We should attempt to make this a roughly conforming oauth connector:
- an attrAthentication field
- either user/pass and oauth, or just oauth only (easiest): we have references if you need.
- correct required-attributes

connector-packager/tests/test_resources/oauth_connector/oauth-config.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
+<pluginOAuthConfig>
+    <dbclass>databricks2</dbclass>
```
rosswbrown:
```suggestion
    <dbclass>test_oauth</dbclass>
```

connector-packager/tests/test_resources/oauth_connector/connectionFields.xml
pvanderknyff: 👍

connector-packager/tests/test_resources/oauth_connector/oauth-config.xml
pvanderknyff: Caught this, but realized that this connector was still validated even with the mistake. Had to add a bit of code to make sure we mark the connector as invalid when this element does not match the class of the rest of the connector, plus a test case to make sure it isn't broken in the future.

rosswbrown: @wangxinlei can confirm, but I believe this also enforced at runtime as well.  No reason not to do it here as well though, thanks.

wangxinlei: Nice catch, currently it is not validated against the dbClass in manifest. When we register the plugin we already got the dbClass from manifest so I can add logic to verify this at runtime. The only reason this field is not removed is because for non-connector plugin we would only have this single oauth-config file.

connector-packager/tests/test_resources/oauth_connector/connectionResolver.xml
```diff
@@ -0,0 +1,29 @@
...
+            <attr>authentication</attr>
+            <attr>username</attr>
+            <attr>password</attr>
+            <attr>authentication-type</attr>
+            <attr>instanceurl</attr>
+            <attr>ACCESSTOKEN</attr>
+            <attr>REFRESHTOKEN</attr>
+            <attr>access-token-issue-time</attr>
+            <attr>access-token-expires-in</attr>
+            <attr>oauth-client</attr>
```
rosswbrown, line 21:
```suggestion
            <attr>server</attr>
            <attr>port</attr>
            <attr>authentication</attr>
            <attr>username</attr>
            <attr>password</attr>
```
I believe the OAuth requires attributes are now handled internally.  Above should be correct.

connector-packager/tests/test_resources/oauth_connector/dialect.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<dialect name='PostgresMCD'
```
rosswbrown, line 3:
```suggestion
<dialect name='TestOAuthDialect'
```

rosswbrown (approved):

PR #689, opened 2021-01-12
Added new results workbook

tsjoblad:
In combination with gh-pages pr, I'm adding a new version of the TDVT results workbook


tsjoblad:
Docs PR https://github.com/tableau/connector-plugin-sdk/pull/690

lukewrites (approved):

pvanderknyff (changes requested):
Got the following error when trying to open the workbook:

Unable to complete action
There was a problem connecting to the data source "test_results_combined".

```
Would you like to edit the connection information?
File: /Users/tsjoblad/Desktop/new TDVT Workbook/For GitHub (1:11:21)\test_metadata.csv not found. Please edit the connection
```

Shouldn't this be bundled with the workbook?

tsjoblad:
@pvanderknyff 
> Shouldn't this be bundled with the workbook?

No, I changed it from a .twb**x** to a **.twb** -- it makes it easier to refresh the data source and change the connection information. Docs include how to connect.

pvanderknyff:
> @pvanderknyff
> 
> > Shouldn't this be bundled with the workbook?
> 
> No, I changed it from a .twb**x** to a **.twb** -- it makes it easier to refresh the data source and change the connection information. Docs include how to connect.

Well, it still is looking for the test_metadata.csv file on my machine when I open it, so it's clearly something went wrong. I'm not sure what the test_metadata.csv file is being used for, but it's still looking for it and not finding it. (I accept that the test_results_combined.csv file can be absent until the ser adds their own, but the test_metadata.csv is the same for everyone, so should be bundled in, or we should have instructions on how to add it.)

pvanderknyff:
> @pvanderknyff
> 
> Ah, ok I see what you're saying. I'll add a line to include test_metadata.csv in the working dir. Would you mind trying to repro that and see if it picks it up?

Seems to work fine now. Seems a bit more clunky to keep track of a csv file like that, and it means that if they move the workbook it will stop working, but if editing the datasource was causing problems in practice then this is fine. I suppose the can always bundle into a .twbx themselves.

pvanderknyff (approved):

tsjoblad:
Right, I mean the idea is that the workbook, test results, and metadata all live in the tdvt workspace, not in the SDK clone. So they need to be moved anyway.

pvanderknyff:
> Right, I mean the idea is that the workbook, test results, and metadata all live in the tdvt workspace, not in the SDK clone. So they need to be moved anyway.

May as well move that metadata file then, or change the workbook to look for the csv file in the place it is now. Copy/pasting a csv file is kind of a wierd step imo.

PR #704, opened 2021-02-01
[TDVT 2.4.0] Tests for extract-only datasources

lukewrites:
Allows for testing extract-only data sources by testing materializing the Staples and Calcs tables.

- [x] Merges in & closes the tdvt-3.0 branch so some of those changes can be consumed.
- [x] updates the add_ds command to ask if it is an extract-only source. Live connections will have the extract tests auto-added to their ignore list.
- [ ] Time permitting, make those smoke tests logical again. Come to think of it, maybe these tests should be logical too


rosswbrown:
Would we end up factoring out the [calcs_data](https://github.com/tableau/connector-plugin-sdk/blob/master/tdvt/tdvt/exprtests/standard/setup.calcs_data.txt) test into this setup/expected too?

PR #725, opened 2021-02-23
Enforce OAuth instanceurl usage in packager

rosswbrown:
Field 'instanceurl' is only supported when used with 'authentication'='oauth'.
- add check to packager
- add tests for positive and negative case

All tests pass: `python setup.py test`

There might be a more compact way to write this check.  Please suggest if so, thanks!

### DIFF
--- a/connector-packager/connector_packager/xsd_validator.py
+++ b/connector-packager/connector_packager/xsd_validator.py
@@ -163,6 +163,21 @@ def validate_file_specific_rules(file_to_test: ConnectorFile, path_to_file: Path
                                                  " already exists. Cannot have multiple fields with the same name.")
 
                     return False
+                if field_name == 'instanceurl':
+                    used_for_oauth = False
+                    for conditions in child.iter("conditions"):
+                        for condition in conditions.iter("condition"):
+                            if 'field' in condition.attrib:
+                                if condition.attrib['field'] == 'authentication':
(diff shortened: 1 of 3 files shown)


lukewrites (approved):
I think we could make the check more compact, but at the cost of readability. This is clear…well, _Pythonicly clear_ 😄

connector-packager/connector_packager/xsd_validator.py
```diff
@@ -163,6 +163,21 @@ def validate_file_specific_rules(file_to_test: ConnectorFile, path_to_file: Path
                                                  " already exists. Cannot have multiple fields with the same name.")
 
                     return False
+                if field_name == 'instanceurl':
+                    used_for_oauth = False
+                    for conditions in child.iter("conditions"):
+                        for condition in conditions.iter("condition"):
+                            if 'field' in condition.attrib:
```
lukewrites, line 170: I don't like this any better.
```suggestion
                            for conditions in child.iter("conditions"):
                        for condition in conditions.iter("condition"):
                            if not used_for_oauth:
                                used_for_oauth = ('field' in condition.attrib) & \
                                                    (condition.attrib['field'] == 'authentication') & \
                                                    ('value' in condition.attrib) & \
                                                    (condition.attrib['value'] == 'oauth')
```

wangxinlei (approved):

PR #728, opened 2021-02-24
Add fieldnamePostfix doc change, remove column rename workaround

pvanderknyff:
Added the new fieldnamePostfix option to the documentation on logical query format. Removed the calculated field workaround for not being able to change column names, since while it worked in expression tests it didn't work in the logical tests.

### DIFF
--- a/docs/tdvt.md
+++ b/docs/tdvt.md
@@ -462,27 +462,6 @@ The first thing TDVT does is try and connect to your database using the .tds fil
 
 1. Check tabquerytool.exe. This file should be placed in your Tableau bin directory and tdvt/config/tdvt_override.ini should be updated to point at that executable.
 
-__Tests fail with "Reference to undefined field..."__
-The first thing to check is that the column exists in your test database. If it doesn't, you will need to load the data again.
-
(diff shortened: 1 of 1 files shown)


docs/tdvt.md
```diff
@@ -461,27 +462,6 @@ The first thing TDVT does is try and connect to your database using the .tds fil
...
-
-```xml
-<column datatype='real' name='[discount]' role='measure' type='quantitative'>
-    <calculation class='tableau' formula='[discount_]' />
-</column>
-```
-
-The new column will have the same data, and TDVT will be able to find it.
-
 __Boolean data types are not recognized or your database doesn't support them__
```
rosswbrown, line 465: This this section still valid advice then?

pvanderknyff: It is, we have a couple of TDVT tests in teamcity that use the bool workaround and those pass, so the bool workaround does work and there are valid use cases.

I suspect that there's some part of logical query xml that breaks for most of the columns when they are calculated fields for some datatypes, but not for bools. It's likely possible to get the calculated field workaround working for all the types, but it may require changing the logical query xml for the logical tests, which is not a good idea.

rosswbrown (approved):

PR #729, opened 2021-02-25
Don't overwrite higher minimum tableau version

pvanderknyff:
If you add a later minimum version than Tableau detects, we won't overwrite it. Added test.


connector-packager/tests/test_resources/higher_min_version/manifest.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version='1.0' encoding='utf-8' ?>
```
rosswbrown, line 1: Feel free to make a highly simplified, but still valid, connector for test cases like this.  It doesn't actually need to run against anything.

rosswbrown (approved):

connector-packager/connector_packager/jar_jdk_packager.py
```diff
@@ -44,6 +44,10 @@ def get_min_support_version(file_list: List[ConnectorFile]) -> str:
             min_version_tableau = "2020.3"
             reasons.append("Connector uses Connection Dialogs V2, which was added in the 2020.3 release")
 
+    if float(cur_min_version_tableau) > float(min_version_tableau):
```
sxlin, line 47: [The XSD schema allows the version be more than two parts](https://github.com/tableau/connector-plugin-sdk/blob/master/validation/connector_plugin_manifest_latest.xsd#L92). For example, it could be 2020.4.2. 

We can't use float to compare versions.

pvanderknyff: Good catch.

sxlin (approved):

PR #730, opened 2021-02-26
Defect 1140212: [TDVT] [Presto] Intermittent test failures around Date

devanshsoni9:
The change in expected file is needed to enable running Presto TDVT tests in StandardDBs. 
I can create a new expected file if this will break existing customer's plugin test pipeline or nightly test ?? (need guidance on that )

### DIFF
--- a/tdvt/tdvt/exprtests/standard/expected.setup.operator.date.6.txt
+++ b/tdvt/tdvt/exprtests/standard/expected.setup.operator.date.6.txt
@@ -88,19 +88,19 @@
         <value>%null%</value>
       </tuple>
       <tuple>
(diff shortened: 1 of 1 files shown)


rosswbrown:
This change does align with the most common expected result set for that test across all the expected files, however there is some level of risk if another connector does happen to have this same low level of precision around `date0 - datetime0'` results.

pvanderknyff (approved):
Change looks good. Adding a new expected is not a terrible idea if you're worried about failing other tests, though. Hopefully it'd be caught by a standard db proof run before updating the tdvt version in the monolith if that's the case.

PR #735, opened 2021-03-04
Add date formatting section to dialect docs

erademacher:
Added info about format-date-literal and format-datetime-literal. I'm also going to make a monolith change to give these reasonable defaults, since we're no longer recommending DefaultSQLDialect as a base.

### DIFF
--- a/docs/dialect.md
+++ b/docs/dialect.md
@@ -249,6 +249,17 @@ Example:
    <format-string-literal value='Standard' unicode-prefix='_utf16' />
  ```
 
+### Date Literal Support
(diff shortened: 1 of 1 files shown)


docs/dialect.md
```diff
@@ -249,6 +249,17 @@ Example:
    <format-string-literal value='Standard' unicode-prefix='_utf16' />
  ```
 
+### Date Literal Support
```
rosswbrown, line 252: Can you updated the [SQLDialect.tdd](https://github.com/tableau/connector-plugin-sdk/blob/master/samples/components/dialects/SQLDialect.tdd) with the defaults written out in this style from Base/SQLDialect as well.

rosswbrown (approved):

PR #738, opened 2021-03-08
Split calcs_data into calcs_data, calcs_data.date and calcs_data.time

pvanderknyff:
Splits the date test cases into their own separate test. Most of the differences were on date format, so it makes sense to split the tests so that it's clearer why this test fails.

calcs_data goes from 8 expected files to 3
calcs_data.date has 5 expected files
calcs_data.time has 2 expected files

Ran against the following test suites (these were the ones that the original expected files were generated from):
postgres
sqlserver
oracle
exasolution
teradata
sybaseiq
sybase_ase
snowflake

Passes the standard db suite of TDVT tests.


tdvt/tdvt/exprtests/standard/setup.calcs_data.dates.txt
```diff
@@ -0,0 +1,9 @@
+//Test the data in the table.
+date0
+date1
+date2
+date3
+datetime0
+datetime1
+time0
+time1
```
rosswbrown, line 9: Given `time0` and `time1` aren't actively used in tests at the moment would it make sense to split them into a setup.calcs_data.time.txt?  

A partner wanted to be able to narrowly ignore only those due to lack of TIME type support on their database I believe.

pvanderknyff: That makes sense, I'll split them off.

rosswbrown (approved):
LGTM.  Either as a part of this or as follow-up can you ensure the internal artifact gets updated as well to consume this.

PR #741, opened 2021-03-12
Truncate long cmd_output

pvanderknyff:
Forgot to merge https://github.com/tableau/connector-plugin-sdk/pull/495 for about a year. Opening a new pull request since resolving the merge conflicts would be messier. 

Original description:
Had a tdvt results csv file that became very large since the process output was very large (~20,000 characters). Wrote a check that truncates to 512 [changed to 1024 in new PR] characters instead. Open to ideas of what the cutoff point should be, it may be a bit restrictive as of now.

### DIFF
--- a/tdvt/tdvt/tdvt_core.py
+++ b/tdvt/tdvt/tdvt_core.py
@@ -510,7 +510,7 @@ def get_csv_row_data(tds_name, test_name, test_path, test_result, test_case_inde
 
     # Truncate long process outputs to keep size of csv down
     if len(cmd_output) > 1024:
-        cmd_output = cmd_output[:1022] + ".."
+        cmd_output = cmd_output[:1024] + "<output_truncated>"
 
     priority = test_result.test_metadata.get_priority() if test_result and test_result.test_metadata else 'unknown'
     categories = test_result.test_metadata.concat_categories() if test_result and test_result.test_metadata else 'unknown'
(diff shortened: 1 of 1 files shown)


tdvt/tdvt/tdvt_core.py
```diff
@@ -508,6 +508,10 @@ def get_csv_row_data(tds_name, test_name, test_path, test_result, test_case_inde
     if test_result and test_result.test_config:
         test_type = 'logical' if test_result.test_config.logical else 'expression'
 
+    # Truncate long process outputs to keep size of csv down
+    if len(cmd_output) > 1024:
+        cmd_output = cmd_output[:1022] + ".."
```
rosswbrown, line 513:
```suggestion
        cmd_output = cmd_output[:1024] + "<output_truncated>"
```
More explicit and we don't need to get fancy if we overrun our arbitrary limit.  I'm open to something even higher as well if you want, 4096?

pvanderknyff: I think I originally picked '..' because that's how we truncate for other columns, but I agree that `<output_truncated>` is clearer. I'll also bump up the character limit as well, I think we really only should care if it gets ridiculous as in the test that started this, where the csv was so large it made analyzing it almost impossible because every test had gigantic process outputs.

rosswbrown (approved):

PR #744, opened 2021-03-15
Packager warn during validation about using DefaultSQLDialect

jsun28:
The packager warns the partner when they are using DefaultSQLDialect that DefaultSQLDialect isn't recommended.

Manually tested it with empty base, DefaultSQLDialect as base, and regular base, worked as expected.

### DIFF
--- a/connector-packager/connector_packager/xsd_validator.py
+++ b/connector-packager/connector_packager/xsd_validator.py
@@ -17,8 +17,6 @@ PATH_TO_XSD_FILES = Path("../validation").absolute()
 VALID_XML_EXTENSIONS = ['tcd', 'tdr', 'tdd', 'xml']  # These are the file extensions that we will validate
 PLATFORM_FIELD_NAMES = ['server', 'port', 'sslmode', 'authentication', 'username', 'password', 'vendor1', 'vendor2', 'vendor3']
 VENDOR_FIELD_NAME_PREFIX = 'v-'
-DEFAULTSQLDIALECT_AS_BASE_WARNING_MESSAGE = 'Warning: DefaultSQLDialect is not a recommended base to inherit from, ' \
-                                            'please choose base dialect from the recommended base dialect list here: https://tableau.github.io/connector-plugin-sdk/docs/design#choose-a-dialect'
 
 # Holds the mapping between file type and XSD file name
 XSD_DICT = {
@@ -71,7 +69,7 @@ def validate_all_xml(files_list: List[ConnectorFile], folder_path: Path) -> bool
         else:
             xml_violations_found += 1
 
-        check_file_content(file_to_test, path_to_file)
+        warn_file_specific_rules(file_to_test, path_to_file)
 
     if xml_violations_found <= 0:
         logger.debug("No XML violations found")
@@ -182,10 +180,11 @@ def validate_file_specific_rules(file_to_test: ConnectorFile, path_to_file: Path
     return True
 
 # Check if connector file content contains warnings needs to notify connector developer
-def check_file_content(file_to_test: ConnectorFile, path_to_file: Path):
+def warn_file_specific_rules(file_to_test: ConnectorFile, path_to_file: Path):
 
     if file_to_test.file_type == 'dialect':
         xml_tree = parse(str(path_to_file))
         root = xml_tree.getroot()
         if 'base' in root.attrib and root.attrib['base'] == 'DefaultSQLDialect':
-            logger.warning(DEFAULTSQLDIALECT_AS_BASE_WARNING_MESSAGE)
+            logger.warning('Warning: DefaultSQLDialect is not a recommended base to inherit from, ' \
+                           'please see the documentation for current best practices: https://tableau.github.io/connector-plugin-sdk/docs/design#choose-a-dialect')
--- a/connector-packager/connector_packager/xsd_validator.py
+++ b/connector-packager/connector_packager/xsd_validator.py
@@ -69,6 +69,8 @@ def validate_all_xml(files_list: List[ConnectorFile], folder_path: Path) -> bool
         else:
             xml_violations_found += 1
 
+        check_file_content(file_to_test, path_to_file)
+
     if xml_violations_found <= 0:
         logger.debug("No XML violations found")
     else:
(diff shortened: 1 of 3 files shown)


connector-packager/connector_packager/xsd_validator.py
```diff
@@ -69,6 +69,8 @@ def validate_all_xml(files_list: List[ConnectorFile], folder_path: Path) -> bool
         else:
             xml_violations_found += 1
 
+        check_file_content(file_to_test, path_to_file)
```
pvanderknyff, line 72: Do we want to fail the validation if they use DefaultSQLDialect? If so, maybe return a bool from this method and, if it   returns false fail the connector.

connector-packager/connector_packager/xsd_validator.py
```diff
@@ -176,3 +178,13 @@ def validate_file_specific_rules(file_to_test: ConnectorFile, path_to_file: Path
...
+
+# Check if connector file content contains warnings needs to notify connector developer
+def check_file_content(file_to_test: ConnectorFile, path_to_file: Path):
+
+    if file_to_test.file_type == 'dialect':
+        xml_tree = parse(str(path_to_file))
+        root = xml_tree.getroot()
+        if 'base' in root.attrib and root.attrib['base'] == 'DefaultSQLDialect':
+            logger.warning("Warning: DefaultSQLDialect is not a recommended base to inherit from, please change your base dialect to a better one, "
+                           "you can find the recommended base dialect list on our SDK documentation page.")
```
pvanderknyff, line 190: Point to the specific page: https://tableau.github.io/connector-plugin-sdk/docs/design#choose-a-dialect
```suggestion
                           "you can find the recommended base dialect list here: https://tableau.github.io/connector-plugin-sdk/docs/design#choose-a-dialect.")
```

pvanderknyff (commented):
You should write a unit test to test that we correctly flag DefaultSQlDialect, so that it isn't broken accidently in the future.

connector-packager/connector_packager/xsd_validator.py
jsun28: I think we don't want to fail the validation since a warning is preferred here, not an error.

connector-packager/connector_packager/xsd_validator.py
jsun28: Done, thanks!

jsun28:
Unit test added in the second commit.

connector-packager/connector_packager/xsd_validator.py
```diff
@@ -176,3 +180,12 @@ def validate_file_specific_rules(file_to_test: ConnectorFile, path_to_file: Path
                     return False
 
     return True
+
+# Check if connector file content contains warnings needs to notify connector developer
+def check_file_content(file_to_test: ConnectorFile, path_to_file: Path):
+
+    if file_to_test.file_type == 'dialect':
```
rosswbrown, line 187: Would this logic fit within `validate_file_specific_rules`?

jsun28: I tried to add it there when I started this work, decided not to because 1. `validate_file_specific_rules` is validating file, if not valid, will return false, but for this change, we want a warning, if find defaultSQLDialect as base, we still return true
2. `validate_file_specific_rules` is called twice in packaging process, one is at [gererate_file_list](https://github.com/tableau/connector-plugin-sdk/blob/master/connector-packager/connector_packager/package.py#L89), two is at [validate_all_xml](https://github.com/tableau/connector-plugin-sdk/blob/master/connector-packager/connector_packager/package.py#L92), I think this change makes more sense to be called only once in `validate_all_xml`.

connector-packager/tests/test_xsd_validator.py
```diff
@@ -119,3 +119,14 @@ def test_validate_duplicate_fields_absent(self):
...
+
+    def test_warn_defaultSQLDialect_as_base(self):
+
+        test_dialect_file = TEST_FOLDER / "defaultSQLDialect_as_base/dialect.tdd"
+        file_to_test = ConnectorFile("dialect.tdd", "dialect")
+
+        with self.assertLogs('packager_logger', level='WARNING') as cm:
+            check_file_content(file_to_test, test_dialect_file)
+
+        self.assertIn('WARNING:packager_logger:' + DEFAULTSQLDIALECT_AS_BASE_WARNING_MESSAGE, cm.output)
```
rosswbrown, line 131: Maybe just assert you get a warning and it contains "DefaultSQLDialect", that was we don't need to do a shared variable for every message going forward.

connector-packager/connector_packager/xsd_validator.py
rosswbrown: Ok, yeah ideally it would fit within `validate_file_specific_rules`, but I think this is a good compromise for now.

Can you rename this method to `warn_file_specific_rules`

connector-packager/connector_packager/xsd_validator.py
```diff
@@ -17,6 +17,8 @@
 VALID_XML_EXTENSIONS = ['tcd', 'tdr', 'tdd', 'xml']  # These are the file extensions that we will validate
 PLATFORM_FIELD_NAMES = ['server', 'port', 'sslmode', 'authentication', 'username', 'password', 'vendor1', 'vendor2', 'vendor3']
 VENDOR_FIELD_NAME_PREFIX = 'v-'
+DEFAULTSQLDIALECT_AS_BASE_WARNING_MESSAGE = 'Warning: DefaultSQLDialect is not a recommended base to inherit from, ' \
+                                            'please choose base dialect from the recommended base dialect list here: https://tableau.github.io/connector-plugin-sdk/docs/design#choose-a-dialect'
```
rosswbrown, line 21: please
```suggestion
                                            'please choose base dialect from the recommended base dialect list here: https://tableau.github.io/connector-plugin-sdk/docs/design#choose-a-dialect'
```
```suggestion
                                            'please see the documentation for current best practices: https://tableau.github.io/connector-plugin-sdk/docs/design#choose-a-dialect'
```

connector-packager/connector_packager/xsd_validator.py
jsun28: Done, thanks

rosswbrown (approved):

PR #745, opened 2021-03-15
Bump TDVT version

pvanderknyff:
Want to bring the calcs_data test changes into the monolith, so creating a new TDVT release.


rosswbrown (commented):
I think [version.py](https://github.com/tableau/connector-plugin-sdk/blob/master/tdvt/tdvt/version.py) needs to be changed as well

rosswbrown (approved):

PR #763, opened 2021-04-08
Remove internal/deprecated JDBC caps, and add new ones for 2021.2

jkoskela:
Remove internal/deprecated JDBC capabilities, and add new ones.
Most of the new ones are for 2021.2, but I found a couple from earlier versions that were not present.
I'm not sure when CAP_JDBC_METADATA_SUPPRESS_PREPARED_QUERY was added, it's been there for a while.
This is a PR to master, I didn't realize there was a gh-pages branch for 2021.2. Will cherry pick this back later.

### DIFF
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -117,12 +117,7 @@ CAP_JDBC_BIND_DETECT_ALIAS_CASE_FOLDING | Set to 'yes' to allow Tableau to detec
 CAP_JDBC_BIND_SPATIAL_AS_WKB | Set to 'yes' to use WKB for spatial types serialization. Available in Tableau 2020.4 and newer. | &ndash; | &ndash;
 CAP_JDBC_CONVERT_WKB_HEX_STRING | Set to 'yes' to convert a hex string to regular WKB. Available in Tableau 2020.4 and newer. | &ndash; | &ndash;
 CAP_JDBC_EXPORT_BIND_BOOL_AS_INTEGER  | Set to 'yes' to bind Tableau booleans to integer for data insertion. Available in Tableau 2020.2 and newer. | &ndash; | &ndash; 
-CAP_JDBC_EXPORT_DATA_BATCH | Set to 'no' to disable the use of JDBC batch operations for data insert. | yes | yes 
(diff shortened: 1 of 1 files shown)


docs/capabilities.md
```diff
@@ -118,26 +118,31 @@ CAP_JDBC_BIND_SPATIAL_AS_WKB | Set to 'yes' to use WKB for spatial types seriali
...
 CAP_JDBC_METADATA_NUMERIC_DEFAULT_PREC_SCALE_DOUBLE | Set to 'yes' to use precision=17 and no scale for numeric with undefined precision/scale. Available in Tableau 2020.4 and newer. | &ndash; | &ndash;
 CAP_JDBC_METADATA_READ_FOREIGNKEYS | Set to 'no' to disable reading foreign key metadata | yes | yes   
 CAP_JDBC_METADATA_READ_PRIMARYKEYS | Set to 'no' to disable reading primary key metadata | yes | yes 
 CAP_JDBC_METADATA_USE_RESULTSET_FOR_TABLE | Set to 'yes' to get column metadata from the result set of a select * query. Available in Tableau 2020.4 and newer. | &ndash; | &ndash;
+CAP_JDBC_METADATA_SUPPRESS_PREPARED_QUERY | If CAP_JDBC_METADATA_USE_RESULTSET_FOR_TABLE is enabled, set this capability to 'yes' to disable preparing the query used for reading the table metadata. We will execute the query wrapped with a where-false clause. | &ndash; | &ndash;
 CAP_JDBC_PARAMETER_METADATA_REQUIRES_EXECUTE | Set to 'yes' if retrieving parameter metadata for inserts requires execution of an empty batch. Available in Tableau 2020.2 and newer. | &ndash; | &ndash; 
-CAP_JDBC_QUERY_ASYNC | Set to 'yes' to run queries on another thread | &ndash; | **<span style="color:red">yes</span>**
```
rosswbrown, line 133: Seems like CAP_JDBC_QUERY_ASYNC is still unset by default.  Is it just no longer used or intended as internal?

rosswbrown (approved):

docs/capabilities.md
jkoskela: It was only used by JNIProtocol, so it doesn't do anything (correctly). I see that it is used in CanCancelQueries, but it shouldn't be now.  I will open a story to fix that, and add this back to the docs for now. It needs to be set in the plugin manifest for cancellation to work.

PR #792, opened 2021-05-26
Story 1265616: [SDK] Disallow spaces in connection-fields name values

saugatt:
- Add tests for connection field name validation
- Add regex validation for connection field name
- Add tests for attribute-list validation
- Add regex validation for attribute-list
-


validation/connection_fields.xsd
```diff
@@ -25,7 +25,13 @@
     <xs:attribute name="default-value" type="xs:string"/>
     <xs:attribute name="editable" type="xs:boolean"/>
     <xs:attribute name="secure" type="xs:boolean"/>
-    <xs:attribute name="name" type="xs:string" use="required"/>
+    <xs:attribute name="name" use="required">
+     <xs:simpleType>
+        <xs:restriction base="xs:string">
+          <xs:pattern value="[:a-zA-Z_]([a-zA-Z0-9_\-.])*"/>
```
rosswbrown, line 31:
```suggestion
          <xs:pattern value="[:a-zA-Z_]([a-zA-Z0-9_.-])*"/>
```

rosswbrown (changes requested):
Update regex, otherwise changes look good.

connector-packager/tests/test_resources/invalid_field_name_number/connectionFields.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<connection-fields>
+  <field name="server" label="Server" value-type="string" category="endpoint" >
+    <validation-rule reg-exp="^(([a-zA-Z]|[a-zA-Z][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z]|[A-Za-z][A-Za-z0-9\-]*[A-Za-z0-9])$"/>
+  </field>
+
+  <field name="1port" label="Port" value-type="string" category="endpoint" default-value="5432" />
```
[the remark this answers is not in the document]
saugatt: Will update the test files to have fewer fields.

validation/connection_fields.xsd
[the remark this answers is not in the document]
saugatt: Removing `:`
 tried this without the \, but it seems \ is necessary to escape - or the regex would break.

validation/connection_fields.xsd
```diff
@@ -49,6 +55,7 @@
       <xs:enumeration value="selection"/>
       <xs:enumeration value="boolean"/>
       <xs:enumeration value="file"/>
+      <xs:enumeration value="textbox"/>
```
saugatt, line 58: Looks like we were missing the type in packager and only had updated in the monolith.

validation/connection_fields.xsd
```diff
@@ -25,7 +25,13 @@
     <xs:attribute name="default-value" type="xs:string"/>
     <xs:attribute name="editable" type="xs:boolean"/>
     <xs:attribute name="secure" type="xs:boolean"/>
-    <xs:attribute name="name" type="xs:string" use="required"/>
+    <xs:attribute name="name" use="required">
+     <xs:simpleType>
+        <xs:restriction base="xs:string">
+          <xs:pattern value="[a-zA-Z_]([a-zA-Z0-9_.-])*"/>
```
saugatt, line 31: Updated the regex to not require escaping. Looks like we would not need to escape - if it is the last character for the pattern.

validation/tdr_latest.xsd
```diff
@@ -61,7 +61,13 @@
   </xs:simpleType>
   <xs:complexType name="AttributeList-CT">
     <xs:sequence>
-      <xs:element name="attr" maxOccurs="unbounded"/>
```
rosswbrown, line 64: Let's revert this part and let the monolith enforce this during load for now.  I'm not sure we understand all the implications and I don't want to change it on both sides quiet yet.

validation/connection_fields.xsd
rosswbrown, line 31:
```suggestion
          <xs:pattern value="[:a-zA-Z_]([a-zA-Z0-9_.-])*"/>
```
Add back `:` until we know more.  I want this to be able to match the monolith XSD as well.

rosswbrown (changes requested):
Might need to update the tests, but this should be the last bit of changes needed,

validation/tdr_latest.xsd
saugatt: Sure. I'll remove this part.

validation/connection_fields.xsd
saugatt: Adding `:` back.

rosswbrown (approved):

PR #815, opened 2021-07-15
update dialect xsd up to date

jsun28:
Update this dialect xsd as latest near, changes are from near file history 6-10(5 revisions in total)

### DIFF
--- a/validation/tdd_latest.xsd
+++ b/validation/tdd_latest.xsd
@@ -215,8 +215,8 @@ as passed-in parameters.
     <xs:restriction base="xs:string">
       <xs:enumeration value="Case"/>
       <xs:enumeration value="IIF"/>
-      <xs:enumeration value="IIF_"/>
-      <xs:enumeration value="IF"/>
+      <xs:enumeration value="IIF_"/> <!-- added 2021.3 -->
+      <xs:enumeration value="IF"/> <!-- added 2021.3 -->
     </xs:restriction>
   </xs:simpleType>
   <xs:simpleType name="BoolAsValueApproach-ST">
@@ -558,7 +558,7 @@ as passed-in parameters.
       <xs:element maxOccurs="unbounded" minOccurs="0" name="part">
         <xs:complexType>
           <xs:attribute name="name" type="JoinType-ST" use="required"/>
-          <xs:attribute name="value" type="xs:string" />
+          <xs:attribute name="value" type="xs:string" /> <!-- added 2021.3 -->
         </xs:complexType>
       </xs:element>
     </xs:sequence>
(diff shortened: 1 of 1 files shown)


rosswbrown (changes requested):
Some of these will be in 2021.3 and others 2021.4.

Once master has been updated to 2021.2, we'll create 2021.3 and .4 branches here.

jsun28:
> Some of these will be in 2021.3 and others 2021.4.
> 
> Once master has been updated to 2021.2, we'll create 2021.3 and .4 branches here.

Thanks, I changed the target branch to dev-2021.3, and looks like only revision 10 in near is not in dev-2021.3, so removed that change.

validation/tdd_latest.xsd
```diff
@@ -548,6 +558,7 @@ as passed-in parameters.
       <xs:element maxOccurs="unbounded" minOccurs="0" name="part">
         <xs:complexType>
           <xs:attribute name="name" type="JoinType-ST" use="required"/>
+          <xs:attribute name="value" type="xs:string" />
```
rosswbrown, line 561:
```suggestion
          <xs:attribute name="value" type="xs:string" /> <!-- added 2020.3 -->
```

validation/tdd_latest.xsd
```diff
@@ -633,6 +647,7 @@ as passed-in parameters.
       <xs:element minOccurs="0" name="id-max-length" type="ValueUInt-CT"/>
       <xs:element minOccurs="0" name="id-quotes" type="ValueString-CT"/>
       <xs:element minOccurs="0" name="start-of-week-offset" type="ValueUInt-CT"/>
+      <xs:element minOccurs="0" name="start-of-week-format" type="StartOfWeekFormat-CT"/>
```
rosswbrown, line 650:
```suggestion
      <xs:element minOccurs="0" name="start-of-week-format" type="StartOfWeekFormat-CT"/> <!-- added 2021.3 -->
```

validation/tdd_latest.xsd
```diff
@@ -215,6 +215,8 @@ as passed-in parameters.
     <xs:restriction base="xs:string">
       <xs:enumeration value="Case"/>
       <xs:enumeration value="IIF"/>
+      <xs:enumeration value="IIF_"/>
+      <xs:enumeration value="IF"/>
```
rosswbrown, line 219:
```suggestion
      <xs:enumeration value="IIF_"/> <!-- added 2021.3 -->
      <xs:enumeration value="IF"/> <!-- added 2021.3 -->
```

rosswbrown (commented):
I noticed we added this version comment in some other new addition.  Seems handy, we'll see if we can keep it up for now.

validation/tdd_latest.xsd
rosswbrown: This should say 2021.3

sxlin (approved):
Ship it after add the comments Ross suggested.

jsun28:
> I noticed we added this version comment in some other new addition. Seems handy, we'll see if we can keep it up for now.

Added comments as you suggested, and I tested it with packager, validation is working fine.

rosswbrown (approved):

PR #828, opened 2021-07-28
Fix tdvt logical query config loading

jsun28:
This is for TFSID 942383.

Based on this LogicalConfig part of this [doc](https://tableau.github.io/connector-plugin-sdk/docs/tdvt) and tdvt test code, we should care about the value of those options are true or false, but in tdvt code here, it only checks if the key is present or not, without checking the value of it is true or false.

### DIFF
--- a/tdvt/tdvt/config_gen/gentests.py
+++ b/tdvt/tdvt/config_gen/gentests.py
@@ -44,12 +44,12 @@ def get_customized_table_name(attributes, base_table):
 
     t = Template(table_name)
     #These substitution holders are in templates.py.
-    if check_logical_config_boolean_attr_value(attributes,'tablenameUpper'):
+    if check_logical_config_boolean_attr_value(attributes, 'tablenameUpper'):
         table_name = t.substitute(dsName=base_table.upper())
     else:
         table_name = t.substitute(dsName=base_table)
 
-    if check_logical_config_boolean_attr_value(attributes,'tablenameLower'):
+    if check_logical_config_boolean_attr_value(attributes, 'tablenameLower'):
         table_name = table_name.lower()
     elif 'calcsnameLower' in attributes and 'calcs' in table_name.lower():
         table_name = table_name.lower()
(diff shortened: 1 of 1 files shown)


tdvt/tdvt/config_gen/gentests.py
```diff
@@ -37,12 +37,12 @@ def get_customized_table_name(attributes, base_table):
 
     t = Template(table_name)
     #These substitution holders are in templates.py.
-    if 'tablenameUpper' in attributes:
+    if 'tablenameUpper' in attributes and attributes['tablenameUpper'] == 'True':
```
rosswbrown, line 40: Backwards compatibility is probably the biggest consideration here.  Previously any value was considered true.  As written this is the most correct implementation though.

The fewest breaking changes that we only consider 'key' == 'False' or 'key' not specified to be the false cases.  I'm interested in others thoughts as well here.

Either way its worth pulling this out as a helper function so we can make the change once and possibly adjust it over time based on feedback too.

jsun28: Changed this to a helper function.
Backwards compatibility is a good point, looking at [all tdvt ini files ](https://opengrok.prod.tableautools.com/source/search?project=teams_near&full=LogicalConfig&defs=&refs=&path=&type=&xrd=&nn=1&si=full&si=full)have logical config in monolith, they all use the correct format, so they should be good.
so the breaking cases are empty value or any values except for True, I think it is hard to make this change and also backwards compatible since the goal is basically to make the boolean value here works as expected.

tdvt/tdvt/config_gen/gentests.py
```diff
@@ -28,6 +28,13 @@ def get_logical_config_templates(ds_registry):
 def get_logical_config_template(ds_registry, config_name):
     return get_logical_config_templates(ds_registry)[config_name]
 
+def check_logical_config_boolean_attr_value(attrs, key):
+    if key not in attrs:
+        return False
+    if key in attrs and attrs[key] == 'False':
```
rosswbrown, line 34:
```suggestion
    if attrs[key] == 'False':
```
With previous statement we know key will be in attrs.

I missed this in the first round, apologies.  Another idea to help us out going forward.  Add another `if attrs[key] == 'True':` that returns True explicitly.  Then in the fallback case where 'True' is returned for backwards compatibility, we could log something to the console to warn users 'True' and 'False' are the expected values.

jsun28: sounds good, fixed.

rosswbrown (approved):

PR #829, opened 2021-07-29
Enforce connection-normalizer rules in packager

pvanderknyff:
- Sets min_tableau_version to 2021.1 if inferred connection resolver is used
- If no connection-normalizer element and connector uses Connection Dialog V1 (with a .tcd file), reject the connector
- Adds "ConnectorProperties", a new object that holds connector-wide properties that are derived from one file but needed by another file when validating the xml. (Needed to save the uses_tcd bool somewhere)
- Refactors parse_file so we only need to call validate_single_file on a file once for a given connector. Before, we'd run it twice: once to make sure that the xml file won't throw an error in parse_file, and once when running validate_all_xml
- Fixes the valid_connector test resource to use the strings from the resource file so we actually pick up the resource file in the file list
- Adds unit tests to check that we correctly accept/reject connectors using inferred connection resolver based on ConnectionProperties.uses_tcd, that we correctly set the min_tableau_version for a valid connector using inferred connection resolver, and that we correctly set ConnectionProperties.uses_tcd.
- Also fixed assorted typos in test names

### DIFF
--- a/connector-packager/connector_packager/jar_jdk_packager.py
+++ b/connector-packager/connector_packager/jar_jdk_packager.py
@@ -24,7 +24,7 @@ MIN_TABLEAU_VERSION_ATTR = "min-version-tableau"
 TABLEAU_SUPPORT_LINK = "https://www.tableau.com/support"
 
 
-def get_min_support_version(file_list: List[ConnectorFile], cur_min_version_tableau: str, manifest_plugin_elem: ET.Element) -> str:
+def get_min_support_version(file_list: List[ConnectorFile], cur_min_version_tableau: str, manifest_plugin_elem: ET.Element, input_dir: Path) -> str:
     """
     Get the minimum support version based on features used in the connector
 
@@ -57,6 +57,15 @@ def get_min_support_version(file_list: List[ConnectorFile], cur_min_version_tabl
             if 2021.1 > float(min_version_tableau):
                 min_version_tableau = "2021.1"
             reasons.append("Connector uses OAuth, which was added in the 2021.1 release")
+        elif connector_file.file_type == "connector-resolver":
+            # Check to see if we're using inferred connection resolver, which needs 2021.1+
+            tdr_root = ET.parse(input_dir / connector_file.file_name).getroot()
+            attribute_list = tdr_root.find('.//connection-normalizer/required-attributes/attribute-list')
+
+            if not attribute_list:
+                if 2021.1 > float(min_version_tableau):
+                    min_version_tableau = "2021.1"
+                reasons.append("Connector uses inferred connection resolver, which was added in the 2021.1 release")
 
     if version.parse(cur_min_version_tableau) > version.parse(min_version_tableau):
         reasons.append("min-tableau-version set to " + cur_min_version_tableau + ", since that is higher than calculated version of " + min_version_tableau)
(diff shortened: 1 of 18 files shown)


connector-packager/connector_packager/jar_jdk_packager.py
```diff
@@ -57,6 +58,15 @@ def get_min_support_version(file_list: List[ConnectorFile], cur_min_version_tabl
             if 2021.1 > float(min_version_tableau):
                 min_version_tableau = "2021.1"
             reasons.append("Connector uses OAuth, which was added in the 2021.1 release")
+        elif connector_file.file_type == "connection-resolver":
+            # Check to see if we're using inferred connection resolver, which needs 2021.1+
+            tdr_root = ET.parse(input_dir / connector_file.file_name).getroot()
+            attribute_list = tdr_root.find('.//connection-normalizer/required-attributes/attribute-list')
+
+            if not attribute_list:
+                if 2021.1 > float(min_version_tableau):
```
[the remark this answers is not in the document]
pvanderknyff: We don't use float to represent min_version_tableau, it's stored as a string. We only cast to float for the comparisons here. We have a unit test testing a min_tableau_version with a major and minor version number, so that case is supported by the current design: https://github.com/tableau/connector-plugin-sdk/blob/master/connector-packager/tests/test_jar_packager.py#L100

connector-packager/connector_packager/xsd_validator.py
```diff
@@ -145,50 +148,84 @@ def get_xsd_file(file_to_test: ConnectorFile) -> Optional[str]:
...
 
 
-def validate_file_specific_rules(file_to_test: ConnectorFile, path_to_file: Path, xml_violations_buffer: List[str]) -> bool:
+def validate_file_specific_rules(file_to_test: ConnectorFile, path_to_file: Path, xml_violations_buffer: List[str], properties: ConnectorProperties) -> bool:
+    """
+    Arguments:
+        file_to_test {ConnectorFile} -- the file we want to validate
+        path_to_file {Path} -- the path to the file we want to validate
+        xml_violations_buffer {list[str]} -- a list of strings that holds the xml violation messages
+
```
sxlin, line 157: Please add argument `properties`.

sxlin (approved):

connector-packager/connector_packager/xsd_validator.py
pvanderknyff: Added

connector-packager/tests/test_connector_properties.py
```diff
@@ -0,0 +1,47 @@
...
+            ConnectorFile("resources-en_US.xml", "resource")]
+
+        properties_uses_tcd = ConnectorProperties()
+        self.assertTrue(validate_all_xml(files_list, test_folder, properties_uses_tcd), "Valid connector not marked as valid")
+        self.assertTrue(properties_uses_tcd.uses_tcd, "uses_tcd not set to True for connector using .tcd file")
+
+        # Check that validate_all_xml properly sets uses_tcd to False if not using .tcd file
+        test_folder = TEST_FOLDER / Path("modular_dialog_connector")  # This connector uses a conneciton-fields.xml file
+
+        files_list = [
```
rosswbrown, line 36: can you re-use/append to the above one? slightly less copy/paste

rosswbrown (approved):

connector-packager/tests/test_connector_properties.py
pvanderknyff: I could maybe re-use the common files, but there's actually only 2 files in common because some of the file extensions are different, and I feel more comfortable explicitly listing the files each time to prevent test errors that are hard to diagnose.

PR #834, opened 2021-08-09
Packager validates that all fields are in normalizer

pvanderknyff:
- Adds a new ConnectorProperty: connection_fields, which stores the connection fields
- Checks that all fields are in the required attributes list, rejects connector if one is missing
- Adds unit tests for the above scenario

### DIFF
--- a/connector-packager/tests/test_resources/mcd_field_not_in_normalizer/connectionMetadata.xml
+++ b/connector-packager/tests/test_resources/mcd_field_not_in_normalizer/connectionMetadata.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<connection-metadata>
+  <database enabled='true' label='Database'>
+    <field />
+  </database>
+  <schema enabled='false' />
+</connection-metadata>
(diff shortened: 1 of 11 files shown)


connector-packager/tests/test_resources/mcd_field_not_in_normalizer/connectionMetadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<connection-metadata>
+  <database enabled='true' label='Database'>
```
rosswbrown, line 4: I think we do this as a follow-up item, but in many cases database enabled means 'dbname' should also be included as a required attribute.  We'll need to review the use cases I think.

rosswbrown (approved):

connector-packager/tests/test_resources/mcd_field_not_in_normalizer/connectionMetadata.xml
pvanderknyff: Opened TFS 1300614 as follow up

PR #835, opened 2021-08-10
Add o auth to manual test

wangxinlei:
Adding OAuth test to the manual test pass for a plugin connector.
I'm merging to gh-pages since this applies to current version, or basically every version.


docs/manual-test.md
```diff
@@ -179,6 +179,11 @@ The workbook should publish without errors.
         1. In the Publish Data Source dialog box, under **Authentication**, click __Edit__ next to **Refresh not enabled**.
         1. Under **Authentication**, select **Allow refresh access** from the drop-down list.<br/>
         ![]({{ site.baseurl }}/assets/mt-pub-allow-refresh.png)<br/>
+        1. For a datasource using OAuth, go to the server settings page to add your credential for the datasource.
+        ![]({{ site.baseurl }}/assets/oauth-server-addtoken.png)<br/> 
+        1. Then under **Authentication** select **Embed <username>** from the drop-down list.
+        ![]({{ site.baseurl }}/assets/oauth-desktop-publish.png)<br/> 
+
```
pvanderknyff, line 186: Can you separate out the OAuth steps into their own section? So we'd have "General Manual Tests" and "OAuth Manual Tests". That way, when we get the screenshots/video of the results it's easier to find specifically the OAuth test cases.

docs/manual-test.md
```diff
@@ -232,3 +239,16 @@ Create a new workbook on Tableau Server with the connector installed on the serv
...
     **Note:** Web authoring (creating a new connection from the web) is not currently available for all connector superclasses. In those cases, your connector won't appear on the list of connectors on Tableau Server. If you can publish a workbook or data source using your connector to your server, then your connector is loaded correctly, even if you can't see it on the list of connectors.
 
+__Test refresing OAuth token on Tableau Server__
+
+If your connector supports oauth, you need to perform this extra step to make sure Tableau can successfully refresh your token.
+1. Go to user's server settings page and find the pane for Saved Credentials for Data Sources.
+
+1. Find your connector in the connector list and click **Add** button next to it.
+
+1. This will invoke the OAuth flow, go ahead and authenticate and this will save the oauth token securly in Tableau Server.
```
pvanderknyff, line 249:
```suggestion
1. Find your connector in the connector list and click **Add** button next to it. This will invoke the OAuth flow, go ahead and authenticate and this will save the OAuth token securely in Tableau Server.
```

This reads to me like the second point is an explanation of the first point, not a separate action the user is taking. Also fixes typo.

docs/manual-test.md
```diff
@@ -232,3 +239,16 @@ Create a new workbook on Tableau Server with the connector installed on the serv
...
+
+1. Exam the oauth token saved, it should contain a username which uniquely identify you, it may also contain a instanceUrl if your oauthConfig file has OAUTH_SUPPORTS_CUSTOM_DOMAIN enabled.
```
pvanderknyff, line 251:
```suggestion
1. Examine the OAuth token saved, it should contain a username which uniquely identify you. It may also contain a instanceUrl if your oauthConfig file has OAUTH_SUPPORTS_CUSTOM_DOMAIN enabled.
```

pvanderknyff (changes requested):

rosswbrown (commented):
This might not be practical, but ideally the screenshots would be of a sample plugin connector rather than snowflake.

docs/manual-test.md
wangxinlei: Yeah sure I can do that.

docs/manual-test.md
```diff
@@ -232,3 +232,44 @@ Create a new workbook on Tableau Server with the connector installed on the serv
 
     **Note:** Web authoring (creating a new connection from the web) is not currently available for all connector superclasses. In those cases, your connector won't appear on the list of connectors on Tableau Server. If you can publish a workbook or data source using your connector to your server, then your connector is loaded correctly, even if you can't see it on the list of connectors.
 
+## Test an OAuth Connector
+
+If your connector supports OAuth Authentication, besides the previous test steps, there are some extra steps you need to verify for OAuth on Tableau Server.
+
+__prerequisite__
```
pvanderknyff, line 239:
```suggestion
__Prerequisite__
```

docs/manual-test.md
```diff
@@ -232,3 +232,44 @@ Create a new workbook on Tableau Server with the connector installed on the serv
 
     **Note:** Web authoring (creating a new connection from the web) is not currently available for all connector superclasses. In those cases, your connector won't appear on the list of connectors on Tableau Server. If you can publish a workbook or data source using your connector to your server, then your connector is loaded correctly, even if you can't see it on the list of connectors.
 
+## Test an OAuth Connector
+
+If your connector supports OAuth Authentication, besides the previous test steps, there are some extra steps you need to verify for OAuth on Tableau Server.
```
pvanderknyff, line 237: Should we also have a specific Desktop test as well? Just a simple case where they create a workbook on desktop using OAuth. 

While the publish scenario technically covers desktop, it'd be nice to have a desktop-only case that they go through first, so that if there are any OAuth problems they can identify and fix it in a context that is easier to understand and set up. (Also, we can't assume that the generic desktop manual test cases will cover OAuth since in multi-auth connectors people might use a different auth type).

pvanderknyff (commented):
Thanks Eric! Just a couple more things.

docs/manual-test.md
```diff
@@ -232,11 +232,23 @@ Create a new workbook on Tableau Server with the connector installed on the serv
...
+![]({{ site.baseurl }}/assets/oauth-desktop-dialog.png)
+1. A default browser should pop up and you will need authenticate yourself through OAuth, you should see a consent screen like this:(It's a google consent screen for illustration, the actual one you see depends on your Identity Provider)<br/>
+![]({{ site.baseurl }}/assets/oauth-google-consent.png)
+1. After you allow access, you should see this landing page on your browser:
+![]({{ site.baseurl }}/assets/oauth-desktop-complete.png)
+1. Then you can close your browser and you should be able to perform the normal manual test steps [here](#Test-your-connector-with-tableau-desktop)
+
+## Test an OAuth Connector on Tableau Server
 
 If your connector supports OAuth Authentication, besides the previous test steps, there are some extra steps you need to verify for OAuth on Tableau Server.
```
pvanderknyff, line 249: This line now need to be moved before the "Test an OAuth Connector on Tableau Desktop" heading. Also, can you put both "Test an OAuth Connector on Tableau Desktop" and "Test an OAuth Connector on Tableau Server" as sub-headings under a heading like "OAuth Connector Test Cases"? It will be easier to link to the single heading.

pvanderknyff (commented):
Just a couple more nitpicks, then it can go through.

pvanderknyff (approved):
LGTM, thanks for adding this!

tsjoblad:
There are some markdown issues, fyi

PR #837, opened 2021-08-13
remove *ServerCustom elements

jinli-dev:
remove clientIdServerCustom, clientSecretServerCustom, redirectUrisServerCustom for oauthConfig

### DIFF
--- a/validation/oauth_config.xsd
+++ b/validation/oauth_config.xsd
@@ -30,10 +30,6 @@
       <!--Some oauth provider allow native app using oauth without clientSecret-->
       <xs:element type="xs:string" name="clientSecretDesktop" maxOccurs="1" minOccurs="0"/>
       <xs:element type="xs:string" name="redirectUrisDesktop" maxOccurs="unbounded" minOccurs="0"/>
-      <!--Allow registering the client info on tableau server-->
-      <xs:element type="xs:string" name="clientIdServerCustom" maxOccurs="1" minOccurs="0"/>
-      <xs:element type="xs:string" name="clientSecretServerCustom" maxOccurs="1" minOccurs="0"/>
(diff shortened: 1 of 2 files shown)


validation/oauth_config.xsd
```diff
@@ -30,10 +30,6 @@
       <!--Some oauth provider allow native app using oauth without clientSecret-->
```
wangxinlei, line 30: This is unrelated, but do you wanna add  the oauthConfigId field from the monolith xsd, I believe I did not update this. Or I can open a follow-up story to add that.

jinli-dev: yes

rosswbrown: The `oauthConfigId` change will need to be made in [dev-2021.4](https://github.com/tableau/connector-plugin-sdk/tree/dev-2021.4).

pvanderknyff: I think master may be an appropriate place in this case, since these are being removed and we don't want people using them even before 2021.4. They're undocumented save for this sample, so I don't think there's a use case where we would want someone to use these attributes until the new way forward comes out.

rosswbrown: This change as it appears here in master is fine, but any other change, like adding `oauthConfigId`, should be made in the appropriate release branch.

wangxinlei: Yeah thanks Ross good point. @jinli-dev Let's just do the clientIdServerCustom/clientSecretServerCustom/redirectUrisServerCustom changes here, I'll create another item to add oauthConfigId in 2021.4.

rosswbrown (approved):

pvanderknyff (approved):

PR #838, opened 2021-08-17
Update Postgres samples to use Connection Dialogs v2

pvanderknyff:
Update the two Postgres samples to use Connection Dialogs V2, to align with our best practices. 

Tested manually that these two connectors still work, and made sure that the attributes are properly added to the connection string.


samples/plugins/postgres_jdbc/manifest.xml
```diff
@@ -30,7 +30,8 @@
       <customization name="CAP_QUERY_TIME_REQUIRES_CAST" value="yes"/>
     </customizations>
   </connection-customization>
-  <connection-dialog file='connection-dialog.tcd'/>
-  <connection-resolver file="connectionResolver.tdr"/>
+  <connection-fields file='connectionFields.xml'/>
+  <connection-metadata file='connectionMetadata.xml'/>
+  <connection-resolver file="connectionResolver.xml"/>
```
rosswbrown, line 35: I want us to do this tdr to xml change, but I think we'll need to update the docs more to do it. Tracked in 1072286, but not scheduled currently (we could).

samples/plugins/postgres_jdbc/connectionMetadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<connection-metadata>
+  <database enabled='true' label='Database'>
```
rosswbrown, line 4: The label is optional and this value does match the platform defaults, so I think lets remove it.

samples/plugins/postgres_jdbc/connectionFields.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<connection-fields>
+  <field name="server" label="Server" value-type="string" category="endpoint" >
+    <validation-rule reg-exp="^(([a-zA-Z]|[a-zA-Z][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z]|[A-Za-z][A-Za-z0-9\-]*[A-Za-z0-9])$"/>
+  </field>
+
+  <field name="port" label="Port" value-type="string" category="endpoint" default-value="5432" />
+
```
rosswbrown, line 9: You need a non-editable `authentication` field that has a default value of `auth-user-pass`

pvanderknyff: I wonder if we should add this as a check in the packager. Should be pretty easy to do since we keep track of the connection fields now.

rosswbrown (approved):

samples/plugins/postgres_jdbc/connectionFields.xml
rosswbrown: I thought we did, but apparently not. Yes that's a good idea.

PR #839, opened 2021-08-18
Validate that authentication field exists

pvanderknyff:
Even if there is only one auth mode, we need to have the non-editable field in connection-fields so the default is properly set. Had to update a bunch of test files as well. (Plus, a bunch of whitespace changes since my VSCode is set to trim trailing spaces on saving.)

Also, right now authentication being missing in the connection-normalizer required attributes list is a warning, not an error. Is there a valid case for authentication not being there? I'm wondering if we should make that an error, or if I'm going about this PR wrong and should only warn if authentication is missing from the connection fields.

### DIFF
--- a/connector-packager/connector_packager/connector_properties.py
+++ b/connector-packager/connector_packager/connector_properties.py
@@ -3,3 +3,4 @@ class ConnectorProperties:
     def __init__(self):
         self.uses_tcd = False
         self.connection_fields = []
+        self.backwards_compatibility_mode = False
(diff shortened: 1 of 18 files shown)


pvanderknyff:
Ross pointed out that this might block old connectors that don't have the authentication attribute from packaging. Opened a new PR to add a command line argument that allows the user to ignore the validation and try and package anyway: https://github.com/tableau/connector-plugin-sdk/pull/840

pvanderknyff:
Fun fact: people can't review changes if you don't push the commits. Apologies for the delay.

rosswbrown (approved):

PR #842, opened 2021-08-23
Add French (Canada) to supported languages

jie-d:
Update SDK doc `Localize Your Connector` with newly supported French (Canada).

PR #843, opened 2021-08-23
Add French (Canada) to supported languages

jie-d:
Manual test:
- Confirmed that the resources-fr_CA.xml file can be successfully packaged into a taco file 
- Ran the connector packaged with a resources-fr_CA.xml file and confirmed that the strings are correctly translated after setting the language to French (Canada) in Tableau.

Unit test:
- Added Story 1305320: [Connector SDK] Add unit test to test that resource files are correctly detected

Filed a bug found from testing:
- Defect 1306176: [Connector SDK] Connector without a resources-en_US.xml file for localized strings packages successfully and causes Tableau Desktop to hang on startup

Updating related SDK doc:
- PR - https://github.com/tableau/connector-plugin-sdk/pull/842

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 # Tableau Connector SDK Changelog
+## 2021-08-23
+### Added
+- Add language support for French (Canada)
 ## 2021-01-22
 ### Removed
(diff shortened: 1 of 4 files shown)


pvanderknyff (approved):
LGTM, and thanks for fixing the test connector

rosswbrown (approved):

PR #870, opened 2021-09-30
Break up operator.date tests

lukewrites:
This PR breaks out the `operator.date` tests into separate tests.


lukewrites:
> Will [this](https://github.com/tableau/connector-plugin-sdk/blob/master/tdvt/tdvt/metadata/test_metadata.csv) need to be updated as well after this change?

@rosswbrown I've [updated the file](https://github.com/tableau/connector-plugin-sdk/pull/870/files#diff-bc01546cc2d88f5e2eb5b09529218074dfe3806cdb12b9097e43cb97f25ed618), does that look right?

rosswbrown:
> > Will [this](https://github.com/tableau/connector-plugin-sdk/blob/master/tdvt/tdvt/metadata/test_metadata.csv) need to be updated as well after this change?
> 
> @rosswbrown I've [updated the file](https://github.com/tableau/connector-plugin-sdk/pull/870/files#diff-bc01546cc2d88f5e2eb5b09529218074dfe3806cdb12b9097e43cb97f25ed618), does that look right?

If those tests have values in the test_results_combined.csv Priority, Categories and Functions columns then I think its correct.

rosswbrown (approved):

PR #872, opened 2021-10-18
Update connection_fields.xsd with 2022.1 changes

sxlin:
Update connection_fields.xsd with 2021.1 changes
1. Added name attribute restrictions;
2. Added support for placeholder for string, file and textbox fields;
3. Added textbox type.


rosswbrown:
Should be in the 22.1 branch

PR #885, opened 2021-11-15
TFSID 1326381 : [E2E][Migration] Move Initial SQL tests upsteam

madhurabha:
TFSID 1326381 : [E2E][Migration] Move Initial SQL tests upsteam

We wanted to move our E2E initial sql test to more lower level. Adding required setup and expected files for the changes.


tdvt/tdvt/version.py
```diff
@@ -1 +1 @@
-__version__ = '2.4.0'
+__version__ = '2.5.0b1'
```
rosswbrown, line 1: Should this just be 2.5.0?

tdvt/tdvt/exprtests/initialsql/setup.initialsql.notemptable.int.txt
```diff
@@ -0,0 +1 @@
+RAWSQLAGG_STR("SUM(%1)", [int0])
```
rosswbrown, line 1: What is actually being tested with this test case?  The initial sql set isn't referenced by the test, like it is [here](https://github.com/tableau/connector-plugin-sdk/blob/master/tdvt/tdvt/exprtests/initialsql/setup.initialsql.temptable.txt)

tdvt/tdvt/exprtests/initialsql/setup.initialsql.notemptable.str.txt
```diff
@@ -0,0 +1 @@
+RAWSQLAGG_STR("select str1 from Calcs where str1 = 'BINDER CLIPS'")
```
atalukdar123, line 1: If temp table is not referenced, what are we testing here? The setup files seems unnecessary to me.

tdvt/tdvt/version.py
madhurabha: since this was not major change Luke suggested we name it differently

tdvt/tdvt/exprtests/initialsql/setup.initialsql.notemptable.int.txt
madhurabha: Yes I had same concern. I had tried to add better sql statement in the calcs but we have different issues with individual connections. I will pull back these changes as this is getting covered by other test cases.

tdvt/tdvt/version.py
lukewrites: Since we don't have perms to overwrite packages in Artifactory, I've been packaging and uploading w/a beta tag so they can be tested in a TC run. Once you get a clean TC run it should be packaged/uploaded without the `b`. 😄

PR #886, opened 2021-11-16
Update validaton for instanceURL and dbname

saugatt:
The change omits the validation in the required attribute for instance URL. 
The lack of dbname is given as a warning. 
Testing: 
Manually tested that Dermio connector can be packaged. 
Also added test for validating dermio connector


connector-packager/connector_packager/xsd_validator.py
```diff
@@ -285,14 +285,17 @@ def validate_file_specific_rules_tdr(file_to_test: ConnectorFile, path_to_file:
...
+                    continue
                 if field not in attributes:
                     xml_violations_buffer.append("Attribute '" + field + "' in connection-fields but not in required-attributes list.")
                     return False
 
         if len(attributes) > 0 and properties.connection_metadata_database and not properties.uses_tcd:
             if 'dbname' not in attributes:
                 xml_violations_buffer.append("connection-metadata 'database' enabled but 'dbname' not in required-attributes list.")
-                return False
+
```
pvanderknyff, line 297: Since we're no longer rejecting the connector if dbname is in the wrong place, move this logic to [warn_file_spicific_rules_tdr](https://github.com/tableau/connector-plugin-sdk/blob/master/connector-packager/connector_packager/xsd_validator.py#L343) and use logger.warning rather than adding to xml_violations_buffer to be consistent with other warnings

pvanderknyff (changes requested):
Might also be a good idea to validate that instanceurl is not in the tdr file if OAauth is being used, though that may be out of scope here for the defect.

connector-packager/connector_packager/xsd_validator.py
saugatt: Updated this to the warn_file_spicific_rules_tdr and added a warning to the top.

connector-packager/tests/test_resources/instanceurl/valid-connector/oauthConfig.xml
```diff
@@ -0,0 +1,61 @@
+<pluginOAuthConfig>
+    <!-- Copyright (C) 2017-2019 Dremio Corporation. This file is confidential and private property. -->
```
pvanderknyff, line 2: This file is still has references to Dremio. We have a valid oauth test connector [here](https://github.com/tableau/connector-plugin-sdk/tree/master/connector-packager/tests/test_resources/oauth_connector), use that as a base. You can probably also re-use that one outright.

connector-packager/tests/test_xsd_validator.py
```diff
@@ -158,7 +172,7 @@ def test_warn_authentication_attribute(self):
             logging.getLogger('packager_logger').warning('dummy message')
             warn_file_specific_rules(file_to_test, test_tdr_file)
 
-        self.assertEqual(len(cm.output), 1)
+        self.assertEqual(len(cm.output), 2)
```
pvanderknyff, line 175: Why did these tests change?

saugatt: The error messages include the warning for dbname.

connector-packager/tests/test_resources/instanceurl/valid-connector/oauthConfig.xml
saugatt: Had forgotten to add the updated files which I added later. Will try also try with the generic OAuth connector already present in the repo and replace it with that if it works.

connector-packager/connector_packager/xsd_validator.py
```diff
@@ -358,3 +357,6 @@ def warn_file_specific_rules_tdr(path_to_file: Path):
     if not authentication_attr_exists:
         logger.warning("Warning: 'authentication' attribute is missing from "
                        "<connection-normalizer>/<required-attributes>/<attribute-list> in " + str(path_to_file) + ".")
+                       
+    if 'dbname' not in attribute_list:
```
rosswbrown, line 361: I think we want the same conditions as before, just as a warning instead of an error.

Additionally a bit about why: "Warning: connection-metadata 'database' enabled but 'dbname' is not in required-attributes list. Consider adding it if the value is used in connection-builder or connection-properties scripts"

connector-packager/tests/test_resources/instanceurl/dremio/META-INF/CERTIFIC.SF
```diff
@@ -0,0 +1,27 @@
+Signature-Version: 1.0
```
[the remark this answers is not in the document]
saugatt: Removed this file and now using the oauth connector in the repo.

connector-packager/connector_packager/xsd_validator.py
saugatt: Updated the warning and moved the logic to `validate_file_specific_rules_tdr` since properties is not available in `warn_file_specific_rules_tdr` unless we want to change its parameters

connector-packager/tests/test_xsd_validator.py
```diff
@@ -270,11 +284,6 @@ def test_all_fields_in_normalizer(self):
         self.assertTrue(validate_single_file(file_to_test, test_file, xml_violations_buffer, properties),
                         "Valid connector marked as invalid")
 
-        print("Test that normalizer not containing 'dbname' if connection-metadata database field is used is invalidated")
-        file_to_test = ConnectorFile("connectionResolver.xml", "connection-resolver")
-        test_file = TEST_FOLDER / "database_field_not_in_normalizer/connectionResolver.xml"
-        self.assertFalse(validate_single_file(file_to_test, test_file, xml_violations_buffer, properties),
```
rosswbrown, line 276: Leave the test case, but change the assertion and update the comment above to indicate dbname is optional if unused.

saugatt: Updated the test for dbname validation to be valid when dbname is not present.

rosswbrown (approved):

PR #896, opened 2021-12-01
Convert Staples smoke test from expression to logical test

### LINKED ISSUE #889: Unexpected behavior for Staples table and a custom LogicalConfig
**About You:**
Name: Cassidy Mountjoy
Company: Blockpoint Systems

**Your question:**
When configuring my logical config I was having an issue with converting column names to lower case. I found myself needing to create a custom one, which I defined as  
```
[LogicalConfig]
Name = bsql_logical_query
fieldnameLower_underscore = True
tablename = $dsName
tablePrefix = [testv1].
```
This worked beautifully for the `calcs` table and I'm passing the majority of the tests... As far as the `staples` table, not so much. I can't get past the smoke tests because of a column name mismatch, due to upper case columns not matching my lower case in the database. It seems like my logical config is only being applied to `calcs`?
I only say this because I fixed this issue in calcs by adding the logical config but it didn't fix the issue in staples.

Here's my current smoke test error in `staples`, an upper/lowercase mismatch:
image

My tds file has lower case names as well:
image

Any help or insight into this would be greatly appreciated, thanks!

lukewrites:
Converts the `Staples` smoke test from expression tests to logical. Closes #889.

Test cases are:
`Staples`: Select `Discount` column.

Screen Shot 2021-12-01 at 1 42 31 PM
Screen Shot 2021-12-01 at 1 59 06 PM


rosswbrown:
Lets keep Calcs as an expression test and Staples as logical. It matches existing usage relatively well and gives us a smoke test for each.

lukewrites:
> Lets keep Calcs as an expression test and Staples as logical. It matches existing usage relatively well and gives us a smoke test for each.

Fixed in the above commits.

tdvt/tdvt/config_gen/datasource_list.py
```diff
@@ -166,8 +166,17 @@ def load_test(config, test_dir=get_root_dir()):
     dsconfig = config[datasource_section]
     all_ini_sections.remove(datasource_section)
     config_name = dsconfig['Name']
-    run_time_config = RunTimeTestConfig(dsconfig.getint('TimeoutSeconds', 60 * 60), dsconfig.get('MaxThread', '0'), dsconfig.get('CommandLineOverride', ''), dsconfig.getboolean('RunAsPerf', False))
-    run_time_config.set_tabquery_paths(dsconfig.get('TabQueryPathLinux', ''), dsconfig.get('TabQueryPathMac', ''), dsconfig.get('TabQueryPathx64', ''))
+    run_time_config = RunTimeTestConfig(
```
lukewrites, line 169: Sorry for putting these changes in this PR, the unreadability of these two was just killing me.

tdvt/tdvt/logicaltests/expected/connection_test/expected.setup.staples.connection.test.xml
```diff
@@ -0,0 +1,18 @@
...
+    <sql>
+      SELECT &quot;Staples&quot;.&quot;Ship Priority&quot; AS &quot;Ship Priority&quot;&#10;FROM &quot;Staples&quot;&#10;GROUP BY 1&#10;    </sql>
+    <query-time>
+      122.000&#10;    </query-time>
+    <table>
+      <schema>
+        <column>[Staples.postgres].[Ship Priority]</column>
+      </schema>
+      <tuple>
+        <value>&quot;0&quot;</value>
```
rosswbrown, line 14: The switch from `Discount` to `Ship Priority` doesn't seem like it'll do a very good job of capturing if the data is loaded correctly in the Staples table or not since the only value is `0`.  That was my impression of the intent around these tests?

lukewrites: You're right, I wasn't thinking. Fixed in the commit below.

rosswbrown (approved):

PR #899, opened 2021-12-02
Link MCD docs to authentication page

pvanderknyff:
The MCD list of supported auth modes was missing `auth-integrated`. Instead of keeping two lists in sync, just linking the MCD page to the authentication page.


rosswbrown (approved):
Double check after merge the table renders the deep link correctly, otherwise LGTM.

PR #909, opened 2021-12-15
Merge dev-2021.4 into master

pvanderknyff:
Looks like the hive dialect was the only thing that changed in 2021.4. Unfortunately, these are Jianan's changes and he's no longer on the team, so added some other people to double check that the changes are ok.


rosswbrown (approved):

pvanderknyff:
Jianan wrote these commits while employed at Tableau, so no CLA is required.

PR #912, opened 2022-01-04
Packaged tacos have version in file name

pvanderknyff:
Change the name of the packaged taco to include the version number. So our packaged Postgres JDBC sample would be:
postgres_jdbc-v0.0.0.taco

This will make it easier to see which taco is which version without having to inspect the archive. Many of our partners do this already manually.

Updated our xml parser tests to check that the version is correctly taken from the manifest. We don't have to worry about the case when the connector version is not set when we create the package name, as not having plugin-version is an XML violation and the connector will be rejected before then.


pvanderknyff:
Note: not merging this right now.

I haven't merged that one because right now, when two connectors in the same folder conflict, we chose the one that comes first alphabetically. If we tag on the version number on the end, that means that the oldest version will be loaded first in this scenario, since the taco-v1 comes alphabetically before taco-v2.

We have a story on the backlog to revise the loading logic to load the newest version of the connector we find, and we shouldn't add logic to put the version name into the packager until then.

connector-packager/connector_packager/xml_parser.py
```diff
@@ -245,6 +246,11 @@ def parse_file(self, file_to_parse: ConnectorFile) -> bool:
                                   file_to_parse.file_name)
                     return False
 
+            # Set the connector version
+            if 'plugin-version' in child.attrib:
```
rosswbrown, line 250: Should we skip in the case of '0.0.0' as well?

pvanderknyff: I'm not sure there's much of a reason to, 0.0.0 is most of the time (outside of our samples) a mistake when making a submission so it's fine to make it easy to notice. The taco can always be manually renamed after packaging (even after signing), so if for some reason it's desired to not have the version number it can be removed.

rosswbrown (approved):

PR #917, opened 2022-01-25
Update manual test plan for partners

lukewrites:
First attempt at making manual test passes less onerous for partners.

Removed:
- "Verify that you can duplicate the data source" step
- `Verify that the data source connection properties are correct.`
- "Edit your connection" step - they may not have multiple endpoints to test against.
- "Missing connector" steps

Added:
- connecting with each supported auth mode

### DIFF
--- a/docs/manual-test.md
+++ b/docs/manual-test.md
@@ -114,6 +114,12 @@ Perform the publishing tests listed below. For more information about publishing
 
 * Publish a workbook without an extract to Tableau Server *with the connector* installed on the server.
 The workbook should publish without errors.
+    - Be sure not to embed credentials when you publish the workbook.
+    - Open the published workbook.
+    - Verify that a Sign In dialog opens.<br/>
+    ![]({{ site.baseurl }}/assets/mt-embed-credentials.png)<br/>
+    For more information, see [Set Credentials for Accessing Your Published Data](https://onlinehelp.tableau.com/current/pro/desktop/en-us/publishing_sharing_authentication.htm) in the Tableau Desktop and Web Authoring Help.
+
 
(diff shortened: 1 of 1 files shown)


docs/manual-test.md
```diff
@@ -97,59 +89,6 @@ Right-click the data source, then click **Extract Data**. Verify that you can cr
...
-
-1. Right-click the data source and click **Edit Data Source**.<br/>
-![]({{ site.baseurl }}/assets/mt-edit-data-source.png)
-The worksheet opens in Tableau.
-1. In the left pane, under **Connections**, click the dropdown menu next to the server name and click **Edit Connection**.<br/>
-![]({{ site.baseurl }}/assets/mt-edit-connection.png)
-1. Change something. For example, change the server.<br/>
-![]({{ site.baseurl }}/assets/mt-change-server.png)
-1. After you click **Sign In**, you should see the new server name under **Connections**.<br/>
-![]({{ site.baseurl }}/assets/mt-new-server.png)
```
pvanderknyff, line 112: I think this should be kept, edit connection isn't that onerous and I think there are some scenarios where the connector changes can break it.

pvanderknyff (commented):
Looks a lot better! Have one test case I think should be kept, the rest looks good to me.

docs/manual-test.md
lukewrites: Sounds good. Are there options other than "change the server" we could suggest?

lukewrites: reverted

pvanderknyff: Could change the database, or the user account, or auth method. Any of the fields should suffice.

pvanderknyff (approved):

PR #918, opened 2022-02-02
Add code to have version number in the name of the taco

jainam1995:
Add code to packager to make sure that plugin-version is defined in the connectors files.
Add the version of the taco in the name of the taco.

Working on updating load logic of connectors
**Things To Remember:** 

* **`dev-2021.2`**  
=====

### DIFF
--- a/connector-packager/connector_packager/package.py
+++ b/connector-packager/connector_packager/package.py
@@ -125,7 +125,7 @@ def main():
         return
 
     package_dest_path = Path(args.dest)
-    package_name = xmlparser.class_name + PACKAGED_EXTENSION
+    package_name = xmlparser.class_name + '_' + xmlparser.plugin_version + PACKAGED_EXTENSION
 
     if not jdk_create_jar(path_from_args, files_to_package, package_name, package_dest_path):
         logger.info("Taco packaging failed. Check " + str(log_file) + " for more information.")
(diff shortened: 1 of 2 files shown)


pvanderknyff:
There is already a pull request to do add the version to the taco name: https://github.com/tableau/connector-plugin-sdk/pull/912/files

I haven't merged that one because right now, when two connectors in the same folder conflict, we chose the one that comes first alphabetically. If we tag on the version number on the end, that means that the oldest version will be loaded first in this scenario, since the taco-v1 comes alphabetically before taco-v2.

We have a story on the backlog to revise the loading logic to load the newest version of the connector we find, and we shouldn't add logic to put the version name into the packager until then.

Also, adding a check for plugin-version is redundant, as the XSD already enforces that that attribute is present: https://github.com/tableau/connector-plugin-sdk/blob/master/validation/connector_plugin_manifest_latest.xsd#L156

jainam1995:
> There is already a pull request to do add the version to the taco name: https://github.com/tableau/connector-plugin-sdk/pull/912/files
> 
> I haven't merged that one because right now, when two connectors in the same folder conflict, we chose the one that comes first alphabetically. If we tag on the version number on the end, that means that the oldest version will be loaded first in this scenario, since the taco-v1 comes alphabetically before taco-v2.
> 
> We have a story on the backlog to revise the loading logic to load the newest version of the connector we find, and we shouldn't add logic to put the version name into the packager until then.
> 
> Also, adding a check for plugin-version is redundant, as the XSD already enforces that that attribute is present: https://github.com/tableau/connector-plugin-sdk/blob/master/validation/connector_plugin_manifest_latest.xsd#L156

Working on updating the connector load logic in monolith

PR #919, opened 2022-02-04
Fix bug in custom output dir logic.

lukewrites:
Fix bug in `return_logging_path` function. Unfortunately, Python doesn't have a great way to check for arguments that haven't been set, so I use a try/except block.

Did the following: 
- fix logic
- bump version.py
- update readme and changelog

Testing:
- tdvt action generate
- tdvt run postgres --generate


tdvt/tdvt/tdvt.py
```diff
@@ -542,11 +542,12 @@ def check_if_custom_output_dir_exists(custom_output_dir: str) -> bool:
...
-        return os.path.join(args.custom_output_dir, 'tdvt_log_combined.txt')
-    elif args.custom_output_dir:
-        sys.exit("The specified output directory doesn't exist: %s" % Path(args.custom_output_dir))
-    else:
+    try:
+        if args.custom_output_dir and check_if_custom_output_dir_exists(args.custom_output_dir):
+            return os.path.join(args.custom_output_dir, 'tdvt_log_combined.txt')
+        elif args.custom_output_dir:
+            sys.exit("The specified output directory doesn't exist: %s" % Path(args.custom_output_dir))
+    except AttributeError:
```
pvanderknyff, line 550: Do we want to maybe log something in this case to let the user know there was an error?

Also, from looking at the code it seems to only return the default log directory if we get an AttributeError, while before we would return it if we didn't have args.custom_output_dir. Can you double check this still works without the custom output directory argument?

lukewrites: Thanks, made a couple of changes:
1. used `hasattr()` to maintain original logic
2. added a `pass` to return default logger

pvanderknyff (approved):

PR #922, opened 2022-02-15
We are limiting the ini parsing to the specified datasources.

tabsdavis:
**Things To Remember:** 

* **`dev-2021.2`**  
=====

### DIFF
--- a/tdvt/tdvt/config_gen/datasource_list.py
+++ b/tdvt/tdvt/config_gen/datasource_list.py
@@ -365,20 +365,22 @@ class TestRegistry(object):
         # When the ds is a suite, we assign the datasources in the suite to the ds variable.
         registry_ini_file = get_ini_path_local_first('config/registry', ini_file)
         self.load_registry(registry_ini_file)
-        self.create_ds_list()
 
         # Check if any list items are suites themselves. If so we include the datasources in those as well.
-        suites_list = ['all', 'standard', 'all_passing', 'nightly', 'slow', 'all_bigquery_sql', 'all_bigquery', 'all_redshift', 'perf']
-        for x in self.ds:
-            if x in suites_list:
-                self.create_ds_list()
-
-        print(self.ds)
+        suite_names = self.get_suite_names(registry_ini_file)
 
         # Read all the datasource ini files and load the test configuration.
         ini_files = get_all_ini_files_local_first('config')
 
+        # Evaluate each item in the ds argument to determine if it is a suite.
+        # If an item is a suite, we add the datasources from the suite to the ds list
         for d in self.ds:
+            for suite_name in suite_names:
+                if suite_name == d:
+                    ds_list = self.get_datasources_from_suite(suite_name, registry_ini_file)
+                    for item in ds_list:
+                        self.ds.append(item)
+
             # Include only the ini files that apply to a datasource in the list.
             for f in ini_files:
                 if d == os.path.basename(f).replace(".ini", ""):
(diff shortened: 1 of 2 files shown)


tdvt/tdvt/config_gen/datasource_list.py
```diff
@@ -358,24 +357,27 @@ def load_test(config, test_dir=get_root_dir()):
...
-            # Preserve the case of elements.
-            config.optionxform = str
-            try:
-                config.read(f)
-            except configparser.ParsingError as e:
-                logging.debug(e)
-                continue
-
-            self.add_test(load_test(config))
+        for d in ds:
```
lukewrites, line 367: We need to extract this ini parsing logic into a method and run it after `self.suite_map` is set (I believe in the `load_registry()` method below. You could set a breakpoint at tdvt.py:728 to drop into it from the top.

When set, `suite_map` is a dictionary with k:v pairs that look like:
```python
{
    'all': ['athena', 'aurora', ...],
    'standard': [blah, blah, blah, …],
}
```

Right now if we do `tdvt run standard` it won't parse any ini files because it doesn't know which ones to parse and add.

Also, we want to think about more precise pattern matching because there could be shared strings in ds names (ex: postgres and postgres jdbc, or I guess a worse case would be something like mysql and mysql_broken_dont_use 🤣). Not the end of the world, but we might be able to do a straight `==` after splitting the file name rather than an `in` on the filename.

lukewrites (changes requested):
The unit tests failures are in two places: [one](https://github.com/tableau/connector-plugin-sdk/blob/master/tdvt/test/tdvt_test.py#L735) and [two](https://github.com/tableau/connector-plugin-sdk/blob/master/tdvt/test/tdvt_test.py#L615). 

Let me know if I can help with them, it's been a while since we've done much with the unit tests and I'm not sure of the overall state of the tests.

tdvt/tdvt/config_gen/datasource_list.py
```diff
@@ -358,27 +357,55 @@ def load_test(config, test_dir=get_root_dir()):
...
-            except configparser.ParsingError as e:
-                logging.debug(e)
-                continue
-
-            self.add_test(load_test(config))
 
+        for d in self.ds:
+            # Include only the ini files that apply to a datasource in the list.
+            for f in ini_files:
+                if d == os.path.basename(f).replace(".ini", ""):
```
lukewrites, line 384:
```suggestion
                if d == Path(f).stem:
```

tdvt/tdvt/config_gen/datasource_list.py
```diff
@@ -358,27 +357,55 @@ def load_test(config, test_dir=get_root_dir()):
...
         self.suite_map = {}
+        self.ds = ds
+
+        # When the ds is a suite, we assign the datasources in the suite to the ds variable.
+        registry_ini_file = get_ini_path_local_first('config/registry', ini_file)
+        self.load_registry(registry_ini_file)
+        self.create_ds_list()
+
+        # Check if any list items are suites themselves. If so we include the datasources in those as well.
+        suites_list = ['all', 'standard', 'all_passing', 'nightly', 'slow', 'all_bigquery_sql', 'all_bigquery', 'all_redshift', 'perf']
```
lukewrites, line 371: Since the suites are a Tableau-specific thing, we don't want them hard-baked in the source code, will need to figure out how to do this along with how Python figures out what's in a suite.

tdvt/tdvt/config_gen/datasource_list.py
lukewrites: `PurePath`/`Path` has a bunch of [really useful methods](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem)

PR #929, opened 2022-03-08
TFSID 1341953 TDVT Tests for GBQ  Add correct table prefix for new GBQ connector

cezhang16:
**Things To Remember:** 

* **`dev-2021.2`**  
=====


tdvt/tdvt/config_gen/templates.py
```diff
@@ -197,6 +197,14 @@
                         'bool_underscore' : True
                  },
 
+    #[tactile-pulsar-824].[TestV1].[Calcs]
```
rosswbrown, line 200: Try expressing this section in the `googlebigquery.ini` file instead of adding it to the global list.

```
[Datasource]
Name = googlebigquery
LogicalQueryFormat = TEST_googleBigQuery

...
[LogicalConfig]
Name = TEST_googleBigQuery

tablename = $dsName
fieldnameNoSpace = True
tablePrefix = [tactile-pulsar-824].[TestV1].

```

tdvt/tdvt/exprtests/standard/expected.setup.cast.str.14.txt
```diff
@@ -0,0 +1,163 @@
...
+      </schema>
+      <tuple>
+        <value>%null%</value>
+      </tuple>
+      <tuple>
+        <value>&quot;false&quot;</value>
+      </tuple>
+      <tuple>
+        <value>&quot;true&quot;</value>
+      </tuple>
```
rosswbrown, line 54: I believe this is a bug. I looked at all the other expected files and its by design that casting `STR([bool0])` should be null/0/1

tdvt/tdvt/exprtests/standard/expected.setup.cast.str.14.txt
```diff
@@ -0,0 +1,163 @@
+<results>
+  <test name='STR([int1])'>
+    <sql>
+      SELECT CAST(`Calcs`.`int1` AS STRING) AS `TEMP_Test__2617331766__0_`&#10;FROM `tactile-pulsar-824`.`TestV1`.`Calcs` `Calcs`&#10;GROUP BY 1&#10;    </sql>
+    <query-time>
+      527.000&#10;    </query-time>
```
rosswbrown, line 6: Remove the all the `<sql>` and `<query-time>` elements from the expected file to match the others.

rosswbrown (changes requested):

tdvt/tdvt/exprtests/standard/expected.setup.cast.str.14.txt
cezhang16: Removed

tdvt/tdvt/exprtests/standard/expected.setup.cast.str.14.txt
cezhang16: I added `CAP_QUERY_BOOLEXPR_TO_INTEXPR=true`, but it seems like still getting true/false. Is there any places that i should take a look?

tdvt/tdvt/config_gen/templates.py
cezhang16: Fixed :))

tdvt/tdvt/exprtests/standard/expected.setup.cast.str.14.txt
rosswbrown: I made a note in the other review about where you could change the dialect if you wanted to verify the cause of this. Let's handle it as follow-up since we'll want to fork the dialect.

PR #947, opened 2022-03-31
Add JDBC-only requirement to submission guidelines

julie-campbell-sfdc:
Noted that the Tableau Exchange accepts only JDBC connectors now.

### DIFF
--- a/docs/gallery-submission.md
+++ b/docs/gallery-submission.md
@@ -49,12 +49,11 @@ Join our [Tableau Developer Program](https://www.tableau.com/developer) to gain 
 
 ### Step 2: Build your integration
 
-After you've signed up for our developer program, use the information in this Connector SDK documentation to get started building a connector.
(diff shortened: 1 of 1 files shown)


docs/gallery-submission.md
```diff
@@ -49,12 +49,11 @@ Join our [Tableau Developer Program](https://www.tableau.com/developer) to gain
 
 ### Step 2: Build your integration
 
-After you've signed up for our developer program, use the information in this Connector SDK documentation to get started building a connector.
```
pvanderknyff, line 52: Can you add a link to the bug artifacts page (https://tableau.github.io/connector-plugin-sdk/docs/bug-artifacts) here as well? We're trying to advertise that page a bit more to try and reduce the amount of times we need to respond asking for more artifacts. Thanks!

pvanderknyff (commented):
Changes look good to me, there's one small addition that would be useful I think. Thanks!

pvanderknyff (approved):
Looks good to me

PR #949, opened 2022-04-04
Document identifier mangling in dialect

rosswbrown:


docs/dialect.md
```diff
@@ -317,3 +317,27 @@ Example:
...
+The value for this property can be:
+  - __Upper__
+      Database expects uppercase identifiers
+  - __Lower__
+      Database expects lowercase identifiers
+
+**id-max-length** <br/>
+Restrict the maximum identifier length. There is no limit by default.
+
+Example:
```
pvanderknyff, line 338: Do we have defaults for these values? Should we document them as well?

rosswbrown: Each section above mentions defaults. There aren't any restrictions by default.

pvanderknyff (approved):
LGTM

PR #951, opened 2022-04-12
Document shared taco directory

pvanderknyff:
- Document Shared Taco directory as an option for Server. Label it as our preferred option.
- Properly document minerva directory for old way of installing on Server
- Update the installation templates for Server.

### DIFF
--- a/docs/run-taco.md
+++ b/docs/run-taco.md
@@ -6,11 +6,19 @@ Note: the below is intended for connector developers. Customer-facing documentat
 
 # Load order and class name collisions
 
+## Tableau 2022.2 and above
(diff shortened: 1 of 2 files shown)


docs/run-taco.md
```diff
@@ -6,11 +6,19 @@ Note: the below is intended for connector developers. Customer-facing documentat
 
 # Load order and class name collisions
 
+## Tableau 2022.2 and above
```
rosswbrown, line 9: Can you move this part to the 2022.2 branch

pvanderknyff: Reverted. Will move to 2022.2 branch after this change is merged to master and then to the 2022.2 branch to try and avoid merge conflicts.

PR #969, opened 2022-06-14
Adding regex tests from unittests.

tabsdavis:
Add regex tests to optional regex test suite from unittests.
Add string tests to standard tests using hyper's results as baseline.

PR #979, opened 2022-06-29
Sdavis string tests

tabsdavis:
Added string tests from unittests.


tdvt/tdvt/exprtests/standard/expected.setup.string.endswith.null.txt
```diff
@@ -0,0 +1,30 @@
...
+        <column>[cast_calcs.postgres].[TEMP(Test)(2210614618)(0)]</column>
+      </schema>
+      <tuple>
+        <value>true</value>
+      </tuple>
+    </table>
+  </test>
+  <test name='ENDSWITH(Num0, Null)'>
+    <error>
+      Reference to undefined field Num0.&#10;Perhaps you meant [num0]?&#10;&#10;    </error>
```
lukewrites, line 24: We don't want to check in an error as an expected. If a ds doesn't support this test we can add the test case to the excludes list in the ds's .ini file.

tdvt/tdvt/exprtests/standard/setup.string.endswith.null.txt
```diff
@@ -0,0 +1,3 @@
+ENDSWITH(str2, Null)
+ENDSWITH(str1,'')
+ENDSWITH(Num0, Null)
```
lukewrites, line 3: Some of the col names start with an uppercase letter, others with lowercase. In tests/datasets/TestV1/postgres/load_calcs.sql I see that the cols are all named using lowercase. Were these conventions in the tests on the monolith? You could try changing them to lowercase (in this file and others) and then doing the proof run.

tdvt/tdvt/exprtests/standard/setup.string.find.null.txt
```diff
@@ -0,0 +1,2 @@
+FIND(str1, Null)
+find(Int0, Null)
```
lukewrites, line 2: Same as above - `int0` rather than `Int0`?

tdvt/tdvt/exprtests/standard/setup.string.startswith.null.txt
```diff
@@ -0,0 +1,5 @@
+STARTSWITH(str2, 'o')
+STARTSWITH(Int0, Null)
+STARTSWITH(Num0, Null)
```
lukewrites, line 3: same for the above - if I'm missing something about this let me know

nit: can remove the blank lines below, too

lukewrites (changes requested):
A few changes to make then run another proof and see if it helps.

tdvt/tdvt/exprtests/standard/setup.string.endswith.null.txt
```diff
@@ -0,0 +1,3 @@
+ENDSWITH(str2, Null)
+ENDSWITH(str1,'')
+ENDSWITH(num0, Null)
```
rosswbrown, line 3:
```suggestion
ENDSWITH(str1, Null)
```
I think all of these tests should use str1 (no null values) and str2 (null values). An int or num column is out of scope for the string tests.

Its possible all the same coverage can be achieved in these tests by just running test cases on str2

tdvt/tdvt/exprtests/standard/setup.string.startswith.null.txt
```diff
@@ -0,0 +1,5 @@
+STARTSWITH(str2, 'o')
+STARTSWITH(int0, Null)
+STARTSWITH(num0, Null)
```
rosswbrown, line 3:
```suggestion
STARTSWITH(str2, Null)
```
Please let me know if I'm missing why num0 and int0 were added.

lukewrites:
Closing, these changes are included in #969

PR #986, opened 2022-07-07
Story 1324999: Warn users when server is not present as required attribute

saugatt:
[SDK] Packager to enforce 'server' as a required attribute
- Added the packager to warn users when 'server' is not present in the attribute list 
- Added tests to check this the code functions as intended.


rosswbrown (approved):

saugatt:
> 

Opening up a 2 point story to document 'server' and 'authentication' as a required attribute here.

PR #1000, opened 2022-08-10
Add --generate_expected flag

chrisjunlee:
Add --generate_expected flag

**Things To Remember:** 

* **`dev-2021.2`**  
=====

For tests that don't have expected values generated yet, using --generate_expected flag will generate the expected value files to run the tests against.


rosswbrown:
We need to find a way to message this option should only be used for internal use cases for now. We want to avoid partners generating new expected files that pass, but don't match our expected behavior.

tdvt/tdvt/version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = '2.6.9'
+__version__ = '2.6.10b1'
```
lukewrites, line 1: need to fix version number

PR #1010, opened 2022-09-21
Mbha dev

madhurabha:
Change is related to adding new Datetime tests for different connection types.

Separate test has been created for datetime columns with Null values


tdvt/tdvt/exprtests/standard/expected.setup.cast.str.datetime.1.txt
```diff
@@ -0,0 +1,120 @@
+<results>
```
lukewrites, line 1: Should this file be called `tdvt/tdvt/exprtests/standard/expected.setup.cast.str.datetime.txt`?

tdvt/tdvt/exprtests/standard/expected.setup.cast.str.datetime.1.txt
```diff
@@ -0,0 +1,120 @@
...
+  <test name='DATETIME(&apos;1900-01-01 01:00:00&apos;)'>
+    <table>
+      <schema>
+        <column>[cast_calcs.sqlserver].[TEMP(Test)(3780030237)(0)]</column>
```
lukewrites, line 5: We want to use hyper to generate the base reference, unless there's a reason to use a different ds.

tdvt/tdvt/version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = '2.6.9'
+__version__ = '2.6.11b3'
```
lukewrites, line 1:
```suggestion
__version__ = '2.6.11'
```

before merging

lukewrites (changes requested):
A few small things.

tdvt/tdvt/exprtests/standard/expected.setup.cast.str.datetime.1.txt
madhurabha: We have 2 separate expected files

lukewrites: Thanks I missed that.

Please regenerate `expected.setup.cast.str.datetime.txt` using hyper, then we can keep `expected.setup.cast.str.datetime.1.txt` as-is. (I assume hyper will have the same results.) 

For reference, which ds did we use to generate the second set of expecteds? It's confusing since both expecteds say `cast_calcs.sqlserver`.

madhurabha: I generated expected.setup.cast.str.datetime.txt using hyper. Second set is for datasources where second value could not be converted. It is based on sqlserver.

tdvt/tdvt/exprtests/standard/expected.setup.cast.str.datetime.1.txt
madhurabha: We had to create this expected based on sqlserver as for some of the datasources seconds value is not set after cast

rosswbrown:
LGTM, @lukewrites is final sign-off here.

rosswbrown (approved):

tdvt/tdvt/version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = '2.6.9'
+__version__ = '2.6.11b5'
```
lukewrites, line 1:
```suggestion
__version__ = '2.6.11'
```

lukewrites (approved):
LGTM, I updated the version number for you, @madhurabha

I will merge and upload to artifactory so you can do the perforce steps.

PR #1030, opened 2022-10-31
Update CSV output for PerfLab

lukewrites:
Change for internal consumption, making csv output from run friendly for our perf testing system.

This PR adds the following CLI flags:

`--perf-run`: takes no arguments; used for data sources that do not have RunAsPerf set in their .ini file.
`--iteration`: takes a single integer argument; if unspecified, defaults to 0.

The `--perf-run` flag is complementary to existing logic that looks for a `RunAsPerf` boolean in a datasource .ini file. This means that the following commands will have identical results:

```
$ python -m tdvt.tdvt run postgres_perf --iteration 9
$ python -m tdvt.tdvt run postgres --perf-run --iteration 9
```
and 
```
$ python -m tdvt.tdvt run postgres_perf
$ python -m tdvt.tdvt run postgres --perf-run 
```
This will allow us to reduce the number of ini files we manage internally. 

A TDVT run with --perf-run will change the output of `test_results_combined.csv`. The file will have different headers, be pipe-delimited rather than comma-delimited, and hard-codes certain values in csv results depending on test status.


tdvt/tdvt/tabquery.py
```diff
@@ -38,14 +37,15 @@ def get_max_process_level_of_parallelization(desired_threads):
...
+
 def build_tabquery_command_line(work):
     try:
         sys.path.insert(0, get_extensions_dir())
         from extend_tabquery import TabqueryCommandLineExtension
         sys.path.pop(0)
         tb = TabqueryCommandLineExtension()
         logging.debug("Imported extension extend_tabquery")
-    except:
+    except ImportError:
```
rosswbrown, line 48: Will this more specific catch be an issue, or is it always this type?

tdvt/tdvt/tdvt.py
```diff
@@ -45,43 +46,46 @@ class TestOutputFiles(object):
...
 
     @classmethod
-    def copy_output_file(c, src_name, src_dir):
+    def copy_output_file(cls, src_name, src_dir, csv_dialect):
         src = os.path.join(src_dir, src_name)
         logging.debug("Copying {0} to output".format(src))
         try:
             with open(src, 'r', encoding='utf8') as src_file:
-                reader = csv.DictReader(src_file, dialect='tdvt')
+                reader = csv.DictReader(src_file, dialect=csv_dialect)
```
rosswbrown, line 54: Eventually I think we should consider replacing the term dialect with format within this TDVT usage.

tdvt/tdvt/tdvt_core.py
```diff
@@ -563,48 +588,94 @@ def get_csv_row_data(tds_name, test_name, test_path, test_result: TestResult, te
         error_msg = test_result.saved_error_message if test_result.saved_error_message else error_msg
         error_type = test_result.get_error_type()
 
-    columns = [suite, test_set_name, tds_name, test_name, test_path, str(passed), str(matched_expected),
-               str(diff_count), test_case_name, test_type, priority, categories, functions, cmd_output,
-               str(error_msg), str(error_type), float(case.execution_time), generated_sql, actual_tuples,
-               expected_tuples]
+    if is_perf_run:
```
rosswbrown, line 591: Could these duplicated columns parts be more closely encapsulated with the code in constants.py? tdvt_output.py? I'm not sure if this is practical or helps though.

rosswbrown (approved):

tdvt/tdvt/tabquery.py
lukewrites: It would only be this type of error, thrown on 44.

tdvt/tdvt/tdvt.py
lukewrites: That's a very good point. "Dialect" is consistent with Python, but "format" makes more sense. I can do it in this PR.

tdvt/tdvt/tdvt_core.py
lukewrites: I'll open a follow-up for that. We should do it; it'll require some plumbing work that I will want to test pretty thoroughly.

PR #1034, opened 2022-11-08
TFSID 1470733: [OAuth] [Connector SDK] Update Connector SDK to contain updated version of OAuthConfig.xsd file for validation.

ethan-pyke:
…rl fields

**Things To Remember:** 

* **`dev-2021.2`**  
=====

TFSID 1470733: [OAuth] [Connector SDK] Update Connector SDK to contain updated version of OAuthConfig.xsd file for validation.

### DIFF
--- a/connector-packager/connector_packager/jar_jdk_packager.py
+++ b/connector-packager/connector_packager/jar_jdk_packager.py
@@ -55,7 +55,13 @@ def get_min_support_version(file_list: List[ConnectorFile], cur_min_version_tabl
                 min_version_tableau = "2020.3"
             reasons.append("Connector uses Connection Dialogs V2, which was added in the 2020.3 release")
         elif connector_file.file_type == "oauth-config":
-            if 2021.1 > float(min_version_tableau):
+            # Check to see if we're using oauthConfigId, which needs 2023.1+
+            pluginOAuthConfigRoot = ET.parse(input_dir / connector_file.file_name).getroot()
+            oauthConfigId = pluginOAuthConfigRoot.find('.//oauthConfigId')
+            if (oauthConfigId is not None):
+                min_version_tableau = "2023.1"
+                reasons.append("Connector contains an oauthConfigId field, which was added in the 2023.1 release")
+            elif 2021.1 > float(min_version_tableau):
                 min_version_tableau = "2021.1"
             reasons.append("Connector uses OAuth, which was added in the 2021.1 release")
         elif connector_file.file_type == "connection-resolver":
--- a/connector-packager/connector_packager/jar_jdk_packager.py
+++ b/connector-packager/connector_packager/jar_jdk_packager.py
@@ -55,12 +55,12 @@ def get_min_support_version(file_list: List[ConnectorFile], cur_min_version_tabl
                 min_version_tableau = "2020.3"
             reasons.append("Connector uses Connection Dialogs V2, which was added in the 2020.3 release")
         elif connector_file.file_type == "oauth-config":
-            # Check to see if we're using oauthConfigId, which needs 2023.1+
+            # Check to see if we're using oauthConfigId, which needs 2021.4+
             pluginOAuthConfigRoot = ET.parse(input_dir / connector_file.file_name).getroot()
             oauthConfigId = pluginOAuthConfigRoot.find('.//oauthConfigId')
-            if (oauthConfigId is not None):
-                min_version_tableau = "2023.1"
-                reasons.append("Connector contains an oauthConfigId field, which was added in the 2023.1 release")
+            if (oauthConfigId is not None and 2021.4 > float(min_version_tableau)):
+                min_version_tableau = "2021.4"
+                reasons.append("Connector contains an oauthConfigId field, which was added in the 2021.4 release")
             elif 2021.1 > float(min_version_tableau):
                 min_version_tableau = "2021.1"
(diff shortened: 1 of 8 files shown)


validation/oauth_config.xsd
```diff
@@ -34,6 +35,7 @@
       <xs:element type="xs:string" name="tokenUri"/>
       <xs:element type="xs:string" name="userInfoUri" maxOccurs="1" minOccurs="0"/>
       <xs:element type="xs:string" name="instanceUrlValidationRegex" maxOccurs="1" minOccurs="0"/>
+      <xs:element type="xs:string" name="defaultInstanceUrl" maxOccurs="1" minOccurs="0"/>
```
rosswbrown, line 38: Do we have an item to ensure the [connector-packager](https://github.com/tableau/connector-plugin-sdk/tree/master/connector-packager) sets `min-version-tableau` to 2023.1 if `defaultInstanceUrl` usage is detected?

validation/oauth_config.xsd
```diff
@@ -25,6 +25,7 @@
   <xs:complexType name="pluginOAuthConfigType">
     <xs:sequence>
       <xs:element type="xs:string" name="dbclass"/>
+      <xs:element type="xs:string" name="oauthConfigId" maxOccurs="1" minOccurs="0"/>
```
rosswbrown, line 28: Ensure there is test coverage in the connector-packager XSD validation. You may be able to augment one of the existing [oauth tests](https://github.com/tableau/connector-plugin-sdk/blob/master/connector-packager/tests/test_xsd_validator.py#L124) and its [files](https://github.com/tableau/connector-plugin-sdk/blob/master/connector-packager/tests/test_resources/oauth_connector/oauth-config.xml), or make a new test, depending on the plan for the one below.

validation/oauth_config.xsd
ethan-pyke: I think we are actually going to hold off on adding defaultInstanceUrl for now until we can do more testing on it.  Should I still do this if oauthConfigId is detected?

connector-packager/connector_packager/jar_jdk_packager.py
```diff
@@ -55,7 +55,13 @@ def get_min_support_version(file_list: List[ConnectorFile], cur_min_version_tabl
                 min_version_tableau = "2020.3"
             reasons.append("Connector uses Connection Dialogs V2, which was added in the 2020.3 release")
         elif connector_file.file_type == "oauth-config":
-            if 2021.1 > float(min_version_tableau):
+            # Check to see if we're using oauthConfigId, which needs 2023.1+
+            pluginOAuthConfigRoot = ET.parse(input_dir / connector_file.file_name).getroot()
+            oauthConfigId = pluginOAuthConfigRoot.find('.//oauthConfigId')
+            if (oauthConfigId is not None):
+                min_version_tableau = "2023.1"
```
rosswbrown, line 62: `oauthConfigId` has existed in some form since, XSD at least, since 21.4. Normally the SDK takes the most permissive stance for compatibility. I'm unsure from the context I have what the right choice is here. 

Think about an existing connector with a min-tableau-version of 21.4 and beginning to adopt this feature. Is this a breaking change in context of those older products, or just graceful fallback to default/existing behavior.

pvanderknyff (approved):
LGTM once Ross's comment is addressed. Packager changes look good and well-tested.

connector-packager/connector_packager/jar_jdk_packager.py
pvanderknyff: FWIW, I don't see `oauthConfigId` mentioned in the documentation or the samples on the SDK, and up until the SDK's XSD were change in this PR connectors using it would have been rejected, so I think the odds of a partner connector having this element before 2023.1 would be very small. Still worth confirming that we're not breaking existing connectors, though.

rosswbrown: I was thinking about this within the context of adding functionality to a connector that already exists. I'll defer to the team on what the expected future usage is and if this is right choice.

ethan-pyke: After talking more with my team, we decided to change it to 21.4.  While we are still not sure if it will work at the moment, we decided it's better to keep it loose rather than restricting it to 23.1+.  I believe there will be some follow-up stories to ensure that earlier versions can work.

connector-packager/connector_packager/jar_jdk_packager.py
```diff
@@ -55,7 +55,13 @@ def get_min_support_version(file_list: List[ConnectorFile], cur_min_version_tabl
                 min_version_tableau = "2020.3"
             reasons.append("Connector uses Connection Dialogs V2, which was added in the 2020.3 release")
         elif connector_file.file_type == "oauth-config":
-            if 2021.1 > float(min_version_tableau):
+            # Check to see if we're using oauthConfigId, which needs 2021.4+
+            pluginOAuthConfigRoot = ET.parse(input_dir / connector_file.file_name).getroot()
+            oauthConfigId = pluginOAuthConfigRoot.find('.//oauthConfigId')
+            if (oauthConfigId is not None):
+                min_version_tableau = "2021.4"
```
rosswbrown, line 62: This will work today, but to ensure we don't need to change it in the future, you'll want to check that min_version_tableau isn't already something larger than 2021.4 before making the set call.

rosswbrown (approved):

PR #1057, opened 2023-01-06
Update dialect XSD to include 'DBDATE' in RemoteType-ST

rosswbrown:
- 'DBDATE' value was missing as a valid value in RemoteType-ST
- 'DATE' value was listed twice in RemoteType-ST
- update second 'DATE' to 'DBDATE' to allow support for DBTYPE_DBDATE


pvanderknyff (commented):
Just to double check, replacing DATE with DBDATE is intentional, correct?

pvanderknyff (approved):

PR #1082, opened 2023-02-28
Update connector class documentation

pvanderknyff:
Add mention that class cannot be changed, so we have something to link to in reviews when people try and change the class.


rosswbrown (approved):
I think this might already be mentioned in the gallery document as well.

PR #1100, opened 2023-03-16
Fix XML Elements table

pvanderknyff:
Seems like <br> at the end needs a blank line between it and the table or the whole thing breaks. Also added | to the end of all rows, don't know if that is needed but don't see a reason to revert.

Also, my VSCode is set to remove trailing whitespace on save, so the diff will show I touched other parts of the file.


sidwray (approved):
Ship it!

PR #1101, opened 2023-03-21
Add Custom Schema support and test generation for novel tables

lukewrites:
This PR adds two new features to TDVT:
1. Allows tests to run against db schema other than `TestV1`
2. Allows users to run a subset of TDVT expression tests against their own table.

Documentation for these features is in #1091

This is a large PR that (due to some bad choices on my part) include distractions like whitespace changes. I'm opening it while continuing to refine the docs and adding more test coverage so reviewers have time for discussion. Recommend looking at the code with `Hide whitespace` selected.


tdvt/tdvt/config_gen/gentests.py
```diff
@@ -2,20 +2,21 @@
     Generate datasource specific logical query files based on a set of genericized input files.
 """
 
-import sys
```
lukewrites, line 5: unused import

tdvt/tdvt/config_gen/gentests.py
```diff
@@ -188,124 +204,46 @@ def list_configs(ds_registry, target_config_name=None):
...
-    staples_fields.append('[Ship Charge]'              )
-    staples_fields.append('[Total Cycle Time]'         )
-    staples_fields.append('[Product In Stock]'         )
-    staples_fields.append('[PID]'                      )
-    staples_fields.append('[Market Segment]'           )
-
-    #Go through input and top level subdirs. Create those in the output and then process the files.
+    # TODO: Logic will go here to look at ds_registry and either use calcs/staples or use custom table;
+    #       that table or those tables will be used in process_test_files below.
+
```
lukewrites, line 234: These are moved to `constants.py`. This was originally written with closing `)` aligned; looks nice but doesn't affect anything.

tdvt/test/tool_test/test_generation/full_map_to_calcs.csv
```diff
@@ -0,0 +1,4 @@
+k,n0,n1,n2,n3,n4,s0,s1,s2,s3,i0,i1,i2,i3,b0,b1,b2,b3,d0,d1,d2,d3,t0,t1,dt0,dt1,z
```
rosswbrown, line 1: For tables with lots of columns its going to be easy to get misaligned on these rows. Should we consider json or some other way to group multiple properties for a single column.

lukewrites: That is a much better user experience, I'll refactor to use json.

lukewrites: Updated below.

tdvt/tdvt/setup_env.py
```diff
@@ -27,25 +38,113 @@ def create_tdvt_ini_file():
...
+    """Parse the column names and data types from the csv file."""
+    col_names = []
+    data_types = []
+    for col in custom_table_cols_and_types:
+        col_names.append(col['name'])
+        data_types.append(col['type'])
+    return dict(zip(col_names, data_types))
+
+
+def parse_custom_schema_csv_file(csv_path):
```
rosswbrown, line 60: Can this be removed?

tdvt/tdvt/config_gen/test_creator.py
```diff
@@ -0,0 +1,163 @@
...
+        with open(self.json_file, 'r') as f:
+            user_cols_dict = json.load(f)
+        return user_cols_dict
+
+    def map_user_cols_to_test_args(self) -> Dict[str, List[str]]:
+        user_cols_dict = self._map_user_cols_to_test_cols()
+        test_args_dict = self._parse_col_constants()
+        for user_col in user_cols_dict.keys():
+            for test_col in TEST_ARGUMENT_DATA_TYPES.keys():
+                if user_cols_dict[user_col] == TEST_ARGUMENT_DATA_TYPES[test_col]:
```
rosswbrown, line 87: Should we allow matching on name as well?

tdvt/tdvt/config_gen/test_creator.py
```diff
@@ -0,0 +1,163 @@
...
+            for column in extant_cols:
+                if len(test_args_dict[column]) == 0:
+                    new_line = '// {}  {} has no matching column in the user table.\n'.format(line, column)
+                    ignored_line += 1
+                    return new_line, ignored_line
+                else:
+                    no_match = True
+                    column_at_hand = test_args_dict[column].copy()
+                    while column_at_hand and no_match:
+                        possible_user_col = random.choice(column_at_hand)
```
rosswbrown, line 149: I think we should prefer determinism here. Let's say I had Calcs loaded correctly, but it only differed by Column name and something else. Maybe I have to add a suffix in my system to each column and I don't support booleans. I'm thinking if I mimicked the order of the columns that map to Num in the json I could potentially achieve pretty good mapping to existing expected files..

tdvt/test/tool_test/test_generation/full_map_to_calcs.json
```diff
@@ -0,0 +1,137 @@
+{
+    "k": {
+        "type": "VARCHAR",
+        "data_shape": "no_empties_no_nulls",
+        "alts": "False"
```
rosswbrown, line 5: Just an idea: these could be their own boolean properties: empties, negatives, nulls

lukewrites: That sounds fine to me - so we would wind up with an object like this:

``` json
{
      "type": "FLOAT",
      "empties": "True",
      "negatives": "False",
      "nulls": "False"
},
```
much easier to understand and type

tdvt/tdvt/config_gen/test_creator.py
lukewrites: That makes sense and should be a common use case.

tdvt/tdvt/config_gen/test_creator.py
lukewrites: Good idea, I'll add it.

tdvt/tdvt/setup_env.py
lukewrites: Removed in [09b5058](https://github.com/tableau/connector-plugin-sdk/pull/1101/commits/09b5058187e3153d2c8d3f49f830d91a296293e5)

tdvt/test/tool_test/test_generation/full_map_to_calcs.json
lukewrites: Changed this in [3cd7571](https://github.com/tableau/connector-plugin-sdk/pull/1101/commits/3cd75712d71b12912e47641401ddf32959c37ddb). Also sorted keys alphabetically.

tdvt/tdvt/config_gen/test_creator.py
lukewrites: fixed in [868d2db](https://github.com/tableau/connector-plugin-sdk/pull/1101/commits/868d2dbb7c5cfcd5b75572f792b066d9e93df153)

tdvt/tdvt/config_gen/test_creator.py
lukewrites: On reflection, I'm thinking this should be a separate workflow; if you have 1:1 mapping, we should allow you to give us a json object of the mapping, saving the user from all the other typing. It could also allow them to run all the tests, instead of a subset of the tests like this PR enables. I'd prefer to add that in a small follow-on PR I think.

rosswbrown (approved):

PR #1131, opened 2023-05-30
Allow one-to-one column matching for TDVT tests

lukewrites:
Allows users to provide a JSON file of column mappings for Staples and Calcs tables. TDVT will then rewrite all expression tests to use the user's cols and save the updated expression tests in a custom test directory. The datasource's .ini file automatically points required expression tests to the user's custom test dir.


tdvt/tdvt/config_gen/test_creator.py
```diff
@@ -161,3 +162,44 @@ def _process_line(line: str, test_args_dict: List[Dict[str, List]]) -> Tuple[str
             for k in user_col_test_col_map.keys():
                 line = line.replace(k, user_col_test_col_map[k])
             return line + '\n', ignored_line
+
+    def create_custom_expression_tests_for_renamed_staples_and_calcs_tables(self):
+        with open(self.json_file, 'r') as f:
```
lukewrites, line 167: @rosswbrown and @pvanderknyff what do you think of using a user-provided json object of table mappings? The alternative was using the applicable bits of the logical config. Any preference? This seemed a bit more straight forward.

lukewrites:
I have two commits to push once my account issues get sorted. I'll also add some screenshots above.

pvanderknyff (commented):
Changes LGTM, is there a place we can write a unit test for one-to-one column mapping?

pvanderknyff (approved):

rosswbrown (approved):

PR #1132, opened 2023-06-02
Multiple embedded OAuth configs supported in 23.1 and newer

rosswbrown:


pvanderknyff (commented):
Info: So the old behavior of using 2021.4 as the minamum version for multiple embedded OAuth configs was made in error? Or did something else change? Could there be connectors using this feature in 21.4 that work that this packager change would prevent from happening?

jkoskela (approved):

PR #1167, opened 2023-08-11
Fix callback URLs, Add README

jkoskela:
- The Azure sample was missing callback URLs
- Use a less metal port number
- Add README


pvanderknyff (approved):

lukewrites (approved):
thanks

PR #1168, opened 2023-08-11
Split OAuth docs into dev and general

jkoskela:
I split this up because I will have the general OAuth instructions linked to from the main Tableau docs for the admins to setup custom OAuth config. The admins shouldn’t need to know about plugin things like javascript and manifests, etc.

Also do some cleanup
- Remove references to "external" config to simplify things. We will just call that "custom."
- Other rewording.
- Changed the main example to use fixed IDP URLs since the custom domain is only for embedded, which is only available for devs.

I validated that all the links work. Some redirect to the section in the other page now.

### DIFF
--- a/docs/oauth-dev.md
+++ b/docs/oauth-dev.md
@@ -1,0 +1,121 @@
+---
+title: OAuth for Plugins
+---
+**IMPORTANT:** OAuth for plugins is available in Tableau 2021.1 and newer. Multi-IDP config for plugins is available starting in Tableau 2023.1.
+
+This document explains how to add OAuth to a connector. It's meant for developers. For an explanation of OAuth configuration and usage see [OAuth Config](./oauth.md).
+
+**In this section**
+
+* TOC
+{:toc}
+
+# How to Enable OAuth for a Plugin Connector
+
+First check your database and driver documentation to make sure it supports OAuth. 
+
+To enable OAuth for your connector add an `<oauth-config>` field in the manifest.xml and link to an oauthConfig.xml you created, described below.
+
+For a complete example please refer to https://github.com/tableau/connector-plugin-sdk/tree/master/samples/scenarios/snowflake_oauth.
+
+```xml
+  manifest.xml
+
+  <?xml version='1.0' encoding='utf-8' ?>
+  <connector-plugin class=...>
+    ...
+    <dialect file='dialect.tdd'/>
+    <oauth-config file='oauthConfigPing.xml'/>    <!-- add this to to define your OAuth Configs -->
+  </connector-plugin>
+```
+
+Starting in Tableau 2023.1, you can add multiple OAuth configs, embedded in the plugin. The end users may also provide external/custom OAuth configurations:
+- By installing the config files in the Tableau directory. See [Custom OAuth Config on Desktop](./oauth.md#custom-oauth-configs-on-desktop)
+- By uploading the config for a site level OAuth client. See [Create Site OAuth Client](./oauth.md#create-site-oauth-client-20231)
+
+However in both cases, at least one embedded config is still required. We are currently working to remove this requirement. See [issue 1108](https://github.com/tableau/connector-plugin-sdk/issues/1108).
+
+```xml
+  manifest.xml
+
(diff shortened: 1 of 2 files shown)


docs/oauth.md
```diff
@@ -1,98 +1,28 @@
 ---
-title: OAuth Authentication Support
+title: OAuth Configuration and Usage
 ---
 **IMPORTANT:** OAuth for plugins is available in Tableau 2021.1 and newer. Multi-IDP config for plugins is available starting in Tableau 2023.1.
 
+This document explains how to configure and use OAuth. It's meant for admins and plugin developers. For information on adding OAuth to a connector plugin, see [OAuth Development](./oauth-dev.md).
```
pvanderknyff, line 6: Question: is the SDK documentation the right place for “general” OAuth custom config documentation? Or should that go to the main Tableau documentation? The SDK docs have generally been targeted solely for people developing connector plugins, not for the general Tableau audience. 

You say the following information is for "admins and plugin developers", yet separate out the "OAuth configuration and Usage" with "OAuth for plugins". I argue that the SDK documentation should only have information related to building and supporting connector plugins, anything else should go to the Tableau documentation.

Instead of adding this to the SDK doucmentaiton, maybe we should get it added to the generic OAuth documentation: https://help.tableau.com/current/server/en-us/protected_auth.htm

pvanderknyff (changes requested):

docs/oauth-dev.md
```diff
@@ -0,0 +1,121 @@
...
+
+  <?xml version='1.0' encoding='utf-8' ?>
+  <connector-plugin class=...>
+    ...
+    <dialect file='dialect.tdd'/>
+    <oauth-config file='oauthConfigPing.xml'/>    <!-- add this to to define your OAuth Configs -->
+  </connector-plugin>
+```
+
+Starting in Tableau 2023.1, you can add multiple OAuth configs, embedded in the plugin. The end users may also provide external/custom OAuth configurations:
```
zsun-tableau, line 32: here we can consider linking to ## Multiple Embedded OAuth Configs portion? 

Also this part is sorta duplicated with this part from oauth page? 
```This file may be provided in the following ways:
- Bundled in the connector package (only available to plugin authors)
- Installing on a client machine for Tableau Desktop, Prep, or Bridge
- Installing for a site on Tableau Cloud or Tableau Server
```

and "The end users may also provide custom OAuth configurations: " i think we should remove word "external".

zsun-tableau (approved):

docs/oauth.md
jkoskela: I will submit this and open a follow up story to have the content copied into the main docs. After that is complete we can redirect.

docs/oauth.md
```diff
@@ -1,98 +1,27 @@
...
```
rosswbrown, line 6:
```suggestion
This document explains how to configure and use OAuth. It's meant for admins and plugin developers. For information on adding OAuth to a connector plugin, see [OAuth Development]({{ site.baseurl }}/docs/oauth-dev).
```

docs/oauth.md
rosswbrown, line 1: I think its ok, but we do reference this page in a few [places](https://github.com/tableau/connector-plugin-sdk/blob/gh-pages/docs/auth-modes.md) other places in the SDK documentation.

rosswbrown (commented):
Overall this seems like a good intermediate step until user documentation is complete.

docs/oauth.md
jkoskela: If I change it VsCode doesn't understand it then. I think both are probably okay.

pvanderknyff (commented):
Seems fine to me as intermediate step, once the links in the rest of the docs are updated. May have been more stubborn than right on Friday.

docs/oauth.md
jkoskela: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#relative-links

"The path of the link will be relative to the current file. Links starting with / will be relative to the repository root. You can use all relative link operands, such as ./ and ../."

rosswbrown (approved):

PR #1169, opened 2023-08-14
Add packager support for null OAuth config

pvanderknyff:
- Tacos with null OAuth config are packaged
- min_tableau_version set to 2024.1 (can always update if we backport)
- Adds a new test, confirmed all other tests pass


pvanderknyff:
Closing for now while  try and fix whatever stupid things happend with git to get it in a bad state. I moved the commits from my mac machine to my windows machine since python was messed up on my mac, must have broken something in that process.

PR #1170, opened 2023-08-15
Add support for null oauth config in packager

pvanderknyff:
- Tacos with null OAuth config are packaged
- min_tableau_version set to 2024.1 (can always update if we backport)
- Adds a new test, confirmed all other tests pass


rosswbrown:
Change looks good, but should we target dev-2024.1 (doesn't exist yet) branch instead?

pvanderknyff:
Think it's fine in the main branch since the resultant taco is already blocked on 2024.1. People won't be able to use the changes to make a taco that works on versions not supported by 2024.1. Plus, there was talk of backporting once release freeze is over.

PR #1219, opened 2024-02-05
Add string.space test alternative for Presto

jkoskela:
Add string.space test alternative for Presto


rosswbrown:
Inputs less than 0 should return NULL.

PR #1228, opened 2024-03-26
Update OAuth examples

jkoskela:
Update OAuth examples
- Add example for Azure with IAM IDC which requires a custom scope
- Add more comments
- Switch username mapping to use email
- Add an example with a config label
- Remove invalid response parameter from Azure (access-token-issue-time)

### DIFF
--- a/samples/components/oauth/athena.okta.xml
+++ b/samples/components/oauth/athena.okta.xml
@@ -1,7 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <pluginOAuthConfig>
   <dbclass>athena</dbclass>
+  <!-- For external configs, prefix with "custom_".  -->
+  <!-- For configs embedded in the connector package, don't prefix with "custom_".  -->
   <oauthConfigId>custom_athena_okta</oauthConfigId>
+  <!-- Config label added in 2023.2. Avoid if backwards compatibility is needed. -->
+  <configLabel>Okta</configLabel>
   <clientIdDesktop>$clientID</clientIdDesktop>
   <clientSecretDesktop>$clientSecret</clientSecretDesktop>
   <redirectUrisDesktop>http://localhost:55556/Callback</redirectUrisDesktop>
@@ -33,7 +37,7 @@
     </entry>
     <entry>
       <key>OAUTH_CAP_CLIENT_SECRET_IN_URL_QUERY_PARAM</key>
-      <value>true</value>
+      <value>false</value>
     </entry>
     <entry>
       <key>OAUTH_CAP_SUPPORTS_GET_USERINFO_FROM_ID_TOKEN</key>
@@ -63,7 +67,7 @@
     </entry>
     <entry>
       <key>username</key>
-      <value>preferred_username</value>
+      <value>email</value>
     </entry>
   </accessTokenResponseMaps>
 </pluginOAuthConfig>
--- a/samples/components/oauth/athena.okta.xml
+++ b/samples/components/oauth/athena.okta.xml
@@ -67,7 +67,7 @@
     </entry>
     <entry>
       <key>username</key>
-      <value>email</value>
+      <value>preferred_username</value>
     </entry>
   </accessTokenResponseMaps>
 </pluginOAuthConfig>
--- a/samples/components/oauth/redshift.azure.iam-idc.xml
+++ b/samples/components/oauth/redshift.azure.iam-idc.xml
@@ -58,6 +58,10 @@
       <value>refresh_token</value>
     </entry>
     <entry>
+      <key>access-token-issue-time</key>
+      <value>issued_at</value>
+    </entry>
+    <entry>
       <key>id-token</key>
       <value>id_token</value>
     </entry>
--- a/samples/components/oauth/redshift.azure.iam-idc.xml
+++ b/samples/components/oauth/redshift.azure.iam-idc.xml
@@ -36,7 +36,7 @@
     </entry>
     <entry>
       <key>OAUTH_CAP_CLIENT_SECRET_IN_URL_QUERY_PARAM</key>
-      <value>true</value>
+      <value>false</value>
     </entry>
     <entry>
       <key>OAUTH_CAP_SUPPORTS_GET_USERINFO_FROM_ID_TOKEN</key>
--- a/samples/components/oauth/redshift.azure.iam-idc.xml
+++ b/samples/components/oauth/redshift.azure.iam-idc.xml
@@ -1,13 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <pluginOAuthConfig>
   <dbclass>redshift</dbclass>
-  <!-- For external configs, prefix with "custom_".  -->
-  <!-- For configs embedded in the connector package, don't prefix with "custom_".  -->
-  <oauthConfigId>custom_redshift_azure</oauthConfigId>
-  <!-- Config label added in 2023.2. Avoid if backwards compatibility is needed. -->
-  <configLabel>Azure</configLabel>
-  <clientIdDesktop>$clientID</clientIdDesktop>
-  <clientSecretDesktop>$clientSecret</clientSecretDesktop>
+  <!-- For configs embedded in the connector package, don't prefix with "custom_". For external configs, always prefix with "custom_". -->
+  <oauthConfigId>custom_redshift_azure_iam_idc</oauthConfigId>
+  <clientIdDesktop>${clientID}</clientIdDesktop>
+  <clientSecretDesktop>${clientSecret}</clientSecretDesktop>
   <redirectUrisDesktop>http://localhost:55556/Callback</redirectUrisDesktop>
   <redirectUrisDesktop>http://localhost:55557/Callback</redirectUrisDesktop>
   <redirectUrisDesktop>http://localhost:55558/Callback</redirectUrisDesktop>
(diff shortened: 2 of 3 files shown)


pvanderknyff (approved):
LGTM

samples/components/oauth/redshift.azure.xml
```diff
@@ -34,12 +37,17 @@
     </entry>
     <entry>
       <key>OAUTH_CAP_CLIENT_SECRET_IN_URL_QUERY_PARAM</key>
-      <value>true</value>
```
zsun-tableau, line 37: for azure, i think we should set this as true? client id and client secret should be in url param: https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#request-an-authorization-code

samples/components/oauth/redshift.azure.xml
```diff
@@ -64,7 +68,7 @@
     </entry>
     <entry>
       <key>username</key>
-      <value>preferred_username</value>
```
zsun-tableau, line 67: why we decided to change the mapping here? if we keep scope "profile", we should have preferred_username info

zsun-tableau: we have been using this mapping for other azure connectors: https://gitlab.tableausoftware.com/connectivity/plugins/azure-sqldb/-/blob/master/src/script/oauthConfig.xml?ref_type=heads#L75

jkoskela: For Okta it may be okay to use either. For Azure, in particular for IAM IDC they have to use v1 Azure tokens which don't have preferred_username claim. [https://learn.microsoft.com/en-us/entra/identity-platform/id-token-claims-reference](https://learn.microsoft.com/en-us/entra/identity-platform/id-token-claims-reference)

> The primary username that represents the user. It could be an email address, phone number, or a generic username without a specified format. Its value is mutable and might change over time. Since it's mutable, this value can't be used to make authorization decisions. It can be used for username hints and in human-readable UI as a username. The profile scope is required to receive this claim. Present only in v2.0 tokens.

samples/components/oauth/redshift.azure.xml
jkoskela: I tested it (just the flow would complete, I don't have Azure OAuth setup for Redshift). I don't see that it says it must be in a URL param, I see that it must be URL encoded. There are examples of passing in POST also.

samples/components/oauth/redshift.azure.xml
jkoskela: I can revert it from this one just to show you can use either.

samples/components/oauth/redshift.azure.xml
```diff
@@ -50,10 +58,6 @@
       <key>REFRESHTOKEN</key>
       <value>refresh_token</value>
     </entry>
-    <entry>
-      <key>access-token-issue-time</key>
```
zsun-tableau, line 54: we have to keep the entry here(at least the key here), even if issued_at is not a valid response from azure. b/c the logic in oauth service is we leverage the map keys to decide the response keys from oauth service: https://gitlab.tableausoftware.com/connectivity/oauth-service-impl/-/blob/master/src/main/java/com/tableau/oauth/configs/PluginOAuthConfig.java?ref_type=heads#L215, and we have fallback logic to set default access-token-issue-time: https://gitlab.tableausoftware.com/connectivity/oauth-service-impl/-/blob/master/src/main/java/com/tableau/oauth/requestors/AuthorizationCodeTokenRequest.java#L74.

Hence, if we remove the entry here, oauth service won't return any access-token-issue-time as part of response into monolith, which means it will force refreshing access token everytime when we hit validateorrefreshaccesstoken method: https://sourcegraph.prod.tableautools.com/development/maestro-2018-1-dev/-/blob/modules/connectors/tabconnectorsbase/main/parser/OAuthValidator.cpp?L59(it will fail validate access token immediately).

We should keep the entry here, we can just set the value to empty or something tho

samples/components/oauth/redshift.azure.iam-idc.xml
```diff
@@ -0,0 +1,73 @@
...
+      <value>id_token</value>
+    </entry>
+    <entry>
+      <key>username</key>
+      <value>email</value>
+    </entry>
+    <entry>
+      <key>access-token-expires-in</key>
+      <value>expires_in</value>
+    </entry>
```
zsun-tableau, line 71: same here, access-token-issue-time entry is needed.

samples/components/oauth/redshift.azure.xml
zsun-tableau: yeah i think either way should work, we have been using it in url param for a while: https://gitlab.tableausoftware.com/connectivity/plugins/azure-sqldb/-/blob/master/src/script/oauthConfig.xml?ref_type=heads#L45, but as long as you tested it we should be good i think

samples/components/oauth/athena.okta.xml
```diff
@@ -63,7 +67,7 @@
     </entry>
     <entry>
       <key>username</key>
-      <value>preferred_username</value>
```
zsun-tableau, line 66: for okta, do we wanna revert here as well?

samples/components/oauth/redshift.azure.iam-idc.xml
```diff
@@ -0,0 +1,73 @@
...
+      <key>REFRESHTOKEN</key>
+      <value>refresh_token</value>
+    </entry>
+    <entry>
+      <key>id-token</key>
+      <value>id_token</value>
+    </entry>
+    <entry>
+      <key>username</key>
+      <value>email</value>
```
zsun-tableau, line 66: preferred_username and add <profile> scope as well?

samples/components/oauth/redshift.azure.xml
```diff
@@ -1,20 +1,23 @@
...
   <redirectUrisDesktop>http://localhost:55558/Callback</redirectUrisDesktop>
   <redirectUrisDesktop>http://localhost:55559/Callback</redirectUrisDesktop>
-  <authUri>https://${msUrlBegin}/oauth2/v2.0/authorize</authUri>
-  <tokenUri>https://${msUrlBegin}/oauth2/v2.0/token</tokenUri>
+  <!-- For multitenant apps use the common endpoint, for single tenant apps use the directory specific endpoint. -->
+  <authUri>https://login.microsoftonline.com/${tenant}/oauth2/v2.0/authorize</authUri>
+  <tokenUri>https://login.microsoftonline.com/${tenant}/oauth2/v2.0/token</tokenUri>
   <scopes>openid</scopes>
   <scopes>email</scopes>
-  <scopes>profile</scopes>
```
zsun-tableau, line 15: revert this too? this is needed if you wanna use preferred_username in mapping

samples/components/oauth/redshift.azure.xml
jkoskela: Well that is surprising, the docs say those aren't required. https://tableau.github.io/connector-plugin-sdk/docs/oauth

jkoskela: Well...not that surprising.

zsun-tableau: yeah we need to update the doc - it won't break anything tho if you don't have access-token-issue-time and access-token-expires-in in config, just tedious to force refresh access token always

zsun-tableau: I can create a followup story to update https://tableau.github.io/connector-plugin-sdk/docs/oauth

zsun-tableau: nvmd you just created it, thanks!

jkoskela: Oh actually I guess this is the right docs, but still not there. https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#successful-response-2

zsun-tableau: right, there is no "issued_at" returned from azure side, but what i meant is we have to keep the <key> part to make it work, like this mapping would work as well:

zsun-tableau: yeah right, so this entry would work:
<entry>
      <key>access-token-issue-time</key>
      <value>any-random-value</value>

since there is no issued_at returned at all from azure side

samples/components/oauth/redshift.azure.iam-idc.xml
```diff
@@ -0,0 +1,77 @@
...
+      <key>OAUTH_CAP_PKCE_REQUIRES_CODE_CHALLENGE_METHOD</key>
+      <value>true</value>
+    </entry>
+    <entry>
+      <key>OAUTH_CAP_SUPPORTS_STATE</key>
+      <value>true</value>
+    </entry>
+    <entry>
+      <key>OAUTH_CAP_CLIENT_SECRET_IN_URL_QUERY_PARAM</key>
+      <value>true</value>
```
zsun-tableau, line 39: okay last q from me, should this be consistent with the other azure config for redshift? false?

zsun-tableau (approved):

samples/components/oauth/redshift.azure.iam-idc.xml
jkoskela: No I don't think so. If there are multiple valid configs it's reasonable to show alternatives.

jkoskela: I reverted the change to the existing example.

samples/components/oauth/redshift.azure.xml
jkoskela: Also thanks for catching this!

PR #1237, opened 2024-05-23
update version of setup-python action

lukewrites:
GitHub Actions now has Mac ARM runners! This is how we found out:
1. The version of setup-python we were using doesn't work on Apple Si runners
2. some of our unit tests were borked on Apple Si

This PR updates the action and also fixes the unit tests.

Fixing here


pvanderknyff (approved):
LGTM if tests pass

PR #1241, opened 2024-07-11
Added redshift okta iam idc sample to the oauth sample files

lauradreith:
**Redshift Okta Identity Center XML file :** 

1. To integrate Tableau with Amazon Redshift using IAM Identity Center, you need to use a custom XML. Users can replace the values starting with the $ sign with Okta values.
=====


lauradreith:
signed salesforce CLA

PR #1247, opened 2024-07-22
Fix null oauth version check

pvanderknyff:
I double checked, and turns out we backported this fix to 2023.3, but the packager stamps the minimum version as 2024.1. Corrected this mistake on the issue and now fixing the packager.


pvanderknyff:
Packager test pass locally, looks like some of the github actions are failing for other reasons, unfortunately right after Luke left for vacation.

zsun-tableau (approved):

PR #1264, opened 2024-10-29
jkean-w-16913162-add-token-note

jasonkean:
added new note

**Things To Remember:** 

* **`dev-2021.2`**  
=====

### DIFF
--- a/docs/oauth.md
+++ b/docs/oauth.md
@@ -7,6 +7,9 @@ This document explains how to configure and use OAuth. It's meant for admins and
 
 We have templates for common IDPs in the [samples](https://github.com/tableau/connector-plugin-sdk/tree/master/samples/components/oauth). In general, you should only need to substitute the desktop client ID and secret, and the URL of the IDP. However, there may be additional modifications needed to account for differences such as with PKCE and scopes.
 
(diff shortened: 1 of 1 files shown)


docs/oauth.md
```diff
@@ -7,6 +7,9 @@ This document explains how to configure and use OAuth. It's meant for admins and
 
 We have templates for common IDPs in the [samples](https://github.com/tableau/connector-plugin-sdk/tree/master/samples/components/oauth). In general, you should only need to substitute the desktop client ID and secret, and the URL of the IDP. However, there may be additional modifications needed to account for differences such as with PKCE and scopes.
 
+<div class="alert alert-info"><b>Note</b>: Single use refresh tokens are not supported for OAuth connections to Tableau at this time. In most cases, you can set up your identity provider (such as Okta or Redshift IDC) to use rolling refresh tokens instead. For more information, see your provider’s OAuth documentation. </div>
```
jkoskela, line 10: Suggest: Single use refresh tokens are not support at this time...

Also note that I don't think our customers are using IDC as the main IDP yet (the one that issues the refresh/access tokens that we have on the tableau data connection). When we first discussed IDC with Amazon, IDC was for policies, and maybe to get Amazon STS tokens. But we still run the OAuth flow against something like Okta. Although there could be customers (maybe Amazon themselves) that setup IDC as the main IDP.

jasonkean: So remove "for oauth connections to Tableau" from the first sentence?

